PR #6, opened 2025-03-22
feat(tools): add double-click action and configurable wait timeout

myselfsaurabh:
This PR introduces two enhancements to improve browser automation capabilities:

1. A new `browser_double_click` tool for snapshot mode, enabling double-click interactions with elements using the same accessibility-based approach as existing tools.

2. An enhanced `browser_wait` tool with an optional `maxTime` parameter, giving users more control over maximum wait durations while maintaining the default 10-second cap for backwards compatibility.

Both changes follow existing patterns and coding conventions, making them intuitive additions to the current toolset.


myselfsaurabh:

myselfsaurabh:

src/tools/snapshot.ts
```diff
@@ -112,6 +112,20 @@ export const type: Tool = {
   },
 };
 
+export const doubleClick: Tool = {
```
pavelfeldman, line 115: This looks good do me

src/tools/common.ts
```diff
@@ -95,8 +98,9 @@ export const wait: Tool = {
   },
   handle: async (context, params) => {
     const validatedParams = waitSchema.parse(params);
+    const maxTimeMs = (validatedParams.maxTime || 10) * 1000;
     const page = await context.ensurePage();
-    await page.waitForTimeout(Math.min(10000, validatedParams.time * 1000));
+    await page.waitForTimeout(Math.min(maxTimeMs, validatedParams.time * 1000));
```
pavelfeldman, line 103: waitForTimeout waits for given time, there is no point in taking time and maxTime from the user. The maxTime cap here is to ensure no stalls.

README.md
```diff
@@ -107,6 +107,8 @@ X Y coordinate space, based on the provided screenshot.
 
 The Playwright MCP provides a set of tools for browser automation. Here are all available tools:
 
+**Navigation Commands**
```
pavelfeldman, line 110: Unrelated changes should go into the separate PR.

pavelfeldman (commented):
Thanks for the PR!

README.md
myselfsaurabh: Thanks @pavelfeldman  for reviewing PR. I made changes as per comment.

src/tools/snapshot.ts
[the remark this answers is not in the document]
myselfsaurabh: @pavelfeldman made changes in readme.

myselfsaurabh:
@pavelfeldman Please merge this code if it looks fine

pavelfeldman:
Code looks good! Where can I give it a try? I can't find anything on web that would use double click.

myselfsaurabh:
I have it on my organization's site, which isn't publicly accessible, but you can try it here https://api.jquery.com/dblclick/

myselfsaurabh:
@pavelfeldman  any update on it. if you can merge it, I can raise PR for other changes

pavelfeldman:
Every command we add pollutes the context, so we are trying to be conservative wrt added capabilities. I understand that your site has double click handler, but on the web it is more of a flawed relic from desktop computing. That's why I asked for a real world example. Jquery playground does not stand for the example. Please feel free to file a new issue if there is a healthy set of web properties that would benefit from it. But since it goes explicitly against both Accessibility guidelines and Mobile readiness, it is going to be an uphill battle.

PR #14, opened 2025-03-25
chore: remove extra comma in file `tsconfig.json`

samuel871211:
Why：To make sure it's a valid JSON format


Skn0tt:
Thanks for the PR. `tsconfig.json` allows extraneous commas, just like it allows comments. This is fine.

PR #15, opened 2025-03-25
chore: remove duplicate setting `keyword-spacing` in file `eslint.config.mjs`

samuel871211:


pavelfeldman:
We only accept substantial changes—minor tweaks should be submitted as drive-bys. Thanks.

PR #16, opened 2025-03-25
localStorage tools were added

NebelskyiDmytro:
## Overview
This PR adds comprehensive localStorage handling capabilities to the Playwright MCP repository, enabling automation scripts to interact with browser localStorage. The implementation includes a complete set of tools for managing localStorage data with proper TypeScript typing, validation, and test coverage.

## Features
- **New Tools Added:**
  - `browser_localStorage_get`: Retrieve items from localStorage by key
  - `browser_localStorage_set`: Store items in localStorage
  - `browser_localStorage_remove`: Remove specific items from localStorage
  - `browser_localStorage_clear`: Clear all items from localStorage
  - `browser_localStorage_getAll`: Get all items currently in localStorage

### New Files
- `src/tools/localStorage.ts`: Core implementation of localStorage tools
  - Uses Zod for schema validation and type safety
  - Implements proper error handling
  - Follows existing codebase patterns and conventions

### Modified Files
- `src/program.ts`: 
  - Added localStorage tools to `commonTools` array
  - Tools available in both snapshot and screenshot modes
- `package.json`:
  - Added `zod` dependency
  - Added `@types/commander` dev dependency
  - Added localStorage tools to exports
- `tests/basic.spec.ts`:
  - Added comprehensive test suite for localStorage operations
  - Covers all new tool functionalities

## Example Usage
```typescript
// Set an item
await tools.call('browser_localStorage_set', {
  key: 'user_preferences',
  value: JSON.stringify({ theme: 'dark' })
});

// Get an item
const response = await tools.call('browser_localStorage_get', {
  key: 'user_preferences'
});

// Get all items
const allItems = await tools.call('browser_localStorage_getAll', {});
```

## Dependencies
- Added `zod`: ^3.22.4
  - Used for runtime validation and schema definition
  - Provides TypeScript type inference
  - Generates JSON Schema for API documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,12 +27,14 @@
     "./tools/common": "./lib/tools/common.js",
     "./tools/screenshot": "./lib/tools/screenshot.js",
     "./tools/snapshot": "./lib/tools/snapshot.js",
+    "./tools/localStorage": "./lib/tools/localStorage.js",
     "./package.json": "./package.json"
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.6.1",
     "commander": "^13.1.0",
     "playwright": "1.52.0-alpha-2025-03-21",
+    "zod": "^3.22.4",
     "zod-to-json-schema": "^3.24.4"
   },
   "devDependencies": {
@@ -40,10 +42,11 @@
     "@eslint/js": "^9.19.0",
     "@playwright/test": "1.52.0-alpha-2025-03-21",
     "@stylistic/eslint-plugin": "^3.0.1",
+    "@types/commander": "^2.12.2",
+    "@types/node": "^22.13.10",
     "@typescript-eslint/eslint-plugin": "^8.26.1",
     "@typescript-eslint/parser": "^8.26.1",
     "@typescript-eslint/utils": "^8.26.1",
-    "@types/node": "^22.13.10",
     "eslint": "^9.19.0",
     "eslint-plugin-import": "^2.31.0",
     "eslint-plugin-notice": "^1.0.0",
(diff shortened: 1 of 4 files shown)


NebelskyiDmytro:

pavelfeldman:
Hi there, thanks for the PR. What is your use case for this?

NebelskyiDmytro:
Thanks for the review ! The primary use case for adding localStorage handling tools is to enable stateful automation and faster, more flexible test setups. Here are some key scenarios this unlocks:

- Simulate Authenticated Sessions: Inject auth tokens or user data directly into localStorage to mock login states without relying on backend calls.
- Preserve User Preferences: Easily set or validate user-specific settings like themes, language, or feature flags across different test runs.
- Accelerate Test Setup: Skip repetitive UI flows (e.g., onboarding modals, tutorials) by preloading localStorage with predefined data.
- Improve Debugging: Capture or validate application state during tests by retrieving and inspecting stored values.
- Enable Reusable, Modular Steps: Pass stateful data between test tools or flows using localStorage as a lightweight storage mechanism.

Actually in my case we need to handle state where user is authorized in the browser so we can use LLM models in order to directly access browser with pre-defined state, in order just to avoid repetative actions every time

pavelfeldman:
> Actually in my case we need to handle state where user is authorized in the browser so we can use LLM models in order to directly access browser with pre-defined state, in order just to avoid repetative actions every time

This makes sense, but I have a higher-level idea for it where MCP will hand over control to you to auth once and then it'll remember your storage. Would that work for you?

Issue #21, opened 2025-03-25
Opening any browser with specified storage state


JuAn-Kang:
I've been experiementing with having the agent navigate and analyze a page.
While rough, it was even able to analyze a page, create a basic model, then generate some test scenarios based off it. 

Two things I've noticed:
- There's no way to specify a browser. Since most of my tests run on Chromium, I want to use that but it always defaults to the Chrome app, and is really slow for some reason. It would be nice if there was a way to specify which browser I want explicitly here, and be able to use the same browser/versions as the ones used in test runs. This is similar to #13 .
- It would be nice if there is a way to open with a specific storage state, leveraging the auth json files generated from auth setups. I have several that are specified on a per-testfile basis and am not sure if there's a way to load using one of the storage states.

rajyraman:
@pavelfeldman - Is this happening? Seems like this PR was closed without merging in the changes.

pavelfeldman:
Instead of opening the browser with the specified storage state, we are now using dedicated Chrome profile when opening the browser by default. So all the logged-in state should be preserved. It is not per-test or per-testfile as the primary objective of this MCP is to allow general purpose browser use (for now).

PR #22, opened 2025-03-26
chore: implement saved state

pavelfeldman:
Fixes https://github.com/microsoft/playwright-mcp/issues/21

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -14,57 +14,91 @@
  * limitations under the License.
  */
 
-import * as playwright from 'playwright';
+import fs from 'fs';
+import playwright from 'playwright';
+
+import { storageStateFolder, storageStatePath, storageStatePathIfExists } from './tools/utils';
 
 export class Context {
   private _launchOptions: playwright.LaunchOptions | undefined;
   private _browser: playwright.Browser | undefined;
   private _page: playwright.Page | undefined;
   private _console: playwright.ConsoleMessage[] = [];
-  private _initializePromise: Promise<void> | undefined;
 
   constructor(launchOptions?: playwright.LaunchOptions) {
     this._launchOptions = launchOptions;
   }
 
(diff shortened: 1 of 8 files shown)


src/context.ts
```diff
@@ -14,57 +14,91 @@
...
+      this._browser.on('disconnected', () => {
+        this._browser = undefined;
+        this._reset();
+      });
+    }
+
+    if (!this._page) {
+      const newOrigin = new URL(url).origin;
+      this._page = await this._browser.newPage({
+        storageState: restoreState ? await storageStatePath(newOrigin) : undefined,
```
yury-s, line 55: check if storage state path exists?

yury-s (approved):
Maybe return the storage state to the caller and let it pass it back instead of storing on the filesystem with a chance to leak it between sessions? If we store it on the fs, we should configure an isolated directory per session.

src/tools/utils.ts
```diff
@@ -93,3 +97,65 @@ ${snapshot}
...
+
+type StorageState = Awaited<ReturnType<playwright.BrowserContext['storageState']>>;
+
+export async function saveStorageStateFor(page: playwright.Page, alias: string): Promise<string> {
+  const filePath = await storageStatePath(alias);
+  await page.context().storageState({ path: filePath });
+  return filePath;
+}
+
+export async function loadStorageStateFor(page: playwright.Page, alias: string) {
```
Skn0tt, line 109: just to double check, this and the method above are unused in this PR. is that intended?

pavelfeldman: Oh yes, that was a previous attempt.

pavelfeldman:
> If we store it on the fs, we should configure an isolated directory per session.

The point is that we reuse it between sessions...

rajyraman:
@pavelfeldman - This is going to be great. Much needed to automate things behind Entra

PR #27, opened 2025-03-26
Update README.md

dejely:
Added: Add the Playwright MCP server configuration to your VS Code settings.

For more  clarity of instructions


Skn0tt:
We only accept substantial changes, and I don't think this helps.

PR #42, opened 2025-03-26
Add VS Code install buttons

connor4312:
We set up some redirects so this works now.


pavelfeldman:
How do I test it? (Clicking links did nothing)

connor4312:
The Insiders should link if you're on the latest version of VS Code Insiders. The stable button will work after our release next week.

pavelfeldman:
Remove the stable link to not tease the users for now?

pavelfeldman (approved):

Issue #35, opened 2025-03-26
Question: Iframe interaction


ShaunThayil:
Don't see iframes in the accessibility tree, is there some other way or will this be added afterwards?

Skn0tt:
That's a know limitation of Playwright's aria snapshots, yes. Could you elaborate on your usecase for this? What website do you want to interact with?

ShaunThayil:
I was trying to reserve a spot for an event on eventbrite, but got stuck as the reserve button was in an iframe
```
https://www.eventbrite.com/e/echai-startup-demo-day-in-bengaluru-tickets-1292350538249?aff=ebdssbcitybrowse
```

ShaunThayil:
Other examples would be sites like:
- https://studio.onepeloton.com/new-york/schedule
- https://careers-changegrowlive.icims.com/jobs/

Where the interactible content is within an iframe

Kenn1314:
Hi, may I ask Is it the feature for interacting with iframe has been removed ?

Skn0tt:
Not intentionally, but we've seen bugs with the implementation across multiple pages. If you're experincing issues, please file a bug report and i'll take a look.

PR #49, opened 2025-03-27
fix: stitch all frames into one snapshot

Skn0tt:
Resolves https://github.com/microsoft/playwright-mcp/issues/35 by stitching together aria snapshots for all frames into one. We can't build a proper tree because the ARIA snapshot doesn't contain iframe elements, but concatenating should work just fine for now.

Prompt I used for testing:
```
what jobs are on https://careers-changegrowlive.icims.com/jobs/
```

### DIFF
--- a/src/tools/utils.ts
+++ b/src/tools/utils.ts
@@ -79,15 +79,24 @@ export async function runAndWait(context: Context, status: string, callback: (pa
   };
 }
 
+export async function captureAllFrameSnapshot(page: playwright.Page): Promise<string> {
+  const snapshots = await Promise.all(page.frames().map(frame => frame.locator('html').ariaSnapshot({ ref: true })));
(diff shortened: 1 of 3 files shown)


src/tools/utils.ts
```diff
@@ -79,15 +79,24 @@ export async function runAndWait(context: Context, status: string, callback: (pa
   };
 }
 
+export async function captureAllFrameSnapshot(page: playwright.Page): Promise<string> {
+  const snapshots = await Promise.all(page.frames().map(frame => frame.locator('html').ariaSnapshot({ ref: true })));
```
pavelfeldman, line 83: Let's query all frames via DOM, create frame locators for the nodes that we discover and map frame ids => locator prefixes.

pavelfeldman (approved):

Issue #31, opened 2025-03-26
Upload files is not supported


mahmoudsameer1:
I tried to upload file but. I got this answer 

It seems that the file input field cannot be filled using the standard text input method. Instead, we need to use a method that simulates file selection. Unfortunately, the current setup does not support direct file uploads through the interface I'm using.

PR #52, opened 2025-03-27
feat: browser_choose_file

Skn0tt:
Resolves https://github.com/microsoft/playwright-mcp/issues/31.

Prompt used for testing: 
```
upload test.txt to dropbox
```

This won't work for asynchronously opened filechoosers, but let's start with the synchronous variant. I also tested with including the file chooser open state in the snapshot, but that didn't give better results.

I also tested with OneDrive and WeTransfer, but somehow our ARIA snapshots for those pages are missing some elements that are crucial to locate the upload buttons.


src/tools/utils.ts
```diff
@@ -74,7 +74,7 @@ async function waitForCompletion<R>(page: playwright.Page, callback: () => Promi
 export async function runAndWait(context: Context, status: string, callback: (page: playwright.Page) => Promise<any>, snapshot: boolean = false): Promise<ToolResult> {
   const page = context.existingPage();
   await waitForCompletion(page, () => callback(page));
-  return snapshot ? captureAriaSnapshot(page, status) : {
+  return snapshot ? captureAriaSnapshot(context, page, status) : {
```
pavelfeldman, line 77: I'd clear the file chooser after the action.

pavelfeldman (approved):

PR #60, opened 2025-03-27
chore: only include visible iframes, keep frame locators in own array

Skn0tt:
As discussed:

- hides invisible frames from snapshot
- keep our own frame locator array, so we don't rely on `page.frames()` ordering to be stable

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -101,7 +101,14 @@ export class Context {
       page.locator('html').ariaSnapshot({ ref: true }),
       ...this._lastSnapshotFrames.map(async (frame, index) => {
         const snapshot = await frame.locator('html').ariaSnapshot({ ref: true });
-        return snapshot.replaceAll('[ref=', `[ref=f${index}`);
+        const args = [];
+        const src = await frame.owner().getAttribute('src');
+        if (src)
+          args.push(`src=${src}`);
+        const name = await frame.owner().getAttribute('name');
+        if (name)
+          args.push(`name=${name}`);
+        return `\n# iframe ${args.join(' ')}\n` + snapshot.replaceAll('[ref=', `[ref=f${index}`);
       })
     ]);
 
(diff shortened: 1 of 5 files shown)


src/context.ts
```diff
@@ -90,4 +91,35 @@ export class Context {
...
+  async allFrameSnapshot() {
+    const page = this.existingPage();
+    const visibleFrames = await page.locator('iframe').filter({ visible: true }).all();
+    this._lastSnapshotFrames = visibleFrames.map(frame => frame.contentFrame());
+
+    const snapshots = await Promise.all([
+      page.locator('html').ariaSnapshot({ ref: true }),
+      ...this._lastSnapshotFrames.map(async (frame, index) => {
+        const snapshot = await frame.locator('html').ariaSnapshot({ ref: true });
+        return snapshot.replaceAll('[ref=', `[ref=f${index}`);
```
pavelfeldman, line 104: I would prepend with \n# iframe name=... src=...

src/context.ts
```diff
@@ -90,4 +91,35 @@ export class Context {
     const [page] = context.pages();
     return { page };
   }
+
+  async allFrameSnapshot() {
```
pavelfeldman, line 95: allFramesSnapshot

src/tools/snapshot.ts
```diff
@@ -69,8 +69,8 @@ export const drag: Tool = {
   handle: async (context, params) => {
     const validatedParams = dragSchema.parse(params);
     return runAndWait(context, `Dragged "${validatedParams.startElement}" to "${validatedParams.endElement}"`, async page => {
-      const startLocator = refLocator(page, validatedParams.startRef);
-      const endLocator = refLocator(page, validatedParams.endRef);
+      const startLocator = context.refLocator(validatedParams.startRef);
```
pavelfeldman, line 72: here and in other places, sounds like the page param is no longer used

pavelfeldman (approved):

PR #61, opened 2025-03-27
chore: refactor tests

Skn0tt:


tests/fixtures.ts
```diff
@@ -108,6 +108,49 @@ class MCPServer extends EventEmitter {
       this._child.stdin?.end();
     });
   }
+
+  async listTools() {
+    const list = await this.send({
+      jsonrpc: '2.0',
+      id: 0,
```
pavelfeldman, line 115: Let's increment this to be nice

tests/fixtures.ts
```diff
@@ -108,6 +108,49 @@ class MCPServer extends EventEmitter {
       this._child.stdin?.end();
     });
   }
+
+  async listTools() {
+    const list = await this.send({
```
pavelfeldman, line 113: Your send can now be higher level and accept {method, params} while generating ids and inserting jsonrpc!

pavelfeldman (approved):

PR #62, opened 2025-03-27
developer enhancements - screenshots, console log, and headless mode options

willer:
I know it's 3 features in one PR, but I did these before realizing I should package them up better. They're relatively simple, though, intended for developers who want to get an AI to debug its work:
a) in non-visual mode, allow optional screenshots to be taken (so visual models can avoid using up tokens until a final pass)
b) allow retrieval of console messages, for debugging
c) allow browsing sessions to be started in headless mode, eliminating the need for the startup parameter

(a) and (c) allow optimal choices to be made on the fly rather than at MCP configuration, and (b) is necessary for the AI MCP client to do dev work.


willer:

pavelfeldman:
Let's sort those out one by one. File an issue, explain what you want to see and why. Wait for the positive signal on the issue and proceed with PR!

willer:
seriously? I did the work, and you want me to redo it now?

Issue #67, opened 2025-03-28
ERR_INVALID_URL when using Claude Desktop


medioxor:
i'm getting the following error when trying to use it:
`
browser-1  |     at new URL (node:internal/url:676:13)
browser-1  |     at Server.<anonymous> (/root/.npm/_npx/9833c18b2d85bc59/node_modules/@playwright/mcp/lib/program.js:54:35)                                                                   
browser-1  |     at Server.emit (node:events:517:28)
browser-1  |     at parserOnIncoming (node:_http_server:1131:12)
browser-1  |     at HTTPParser.parserOnHeadersComplete (node:_http_common:119:17) {            
browser-1  |   input: '127.0.0.1:8001/sse?sessionId=e047ce3f-b5bf-467e-b08c-1143ee74d927',     
browser-1  |   code: 'ERR_INVALID_URL'                                                         
browser-1  | }
`

my config is as follows:
`"playwright": {
      "command": "npx",
      "args": [
        "mcp-remote",
        "http://127.0.0.1:8001/sse"
      ]
    }`

medioxor:
fixed it, turns out i have to use localhost rather than 127.0.0.1

Skn0tt:
nope, there's a bug in here!

PR #70, opened 2025-03-28
fix: tired URL parsing bug

Skn0tt:
Closes https://github.com/microsoft/playwright-mcp/issues/67

When `req.headers.host` is defined, the URL parsing breaks because there's no protocol.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -54,9 +54,9 @@ program
       if (options.port) {
         const sessions = new Map<string, SSEServerTransport>();
         const httpServer = http.createServer(async (req, res) => {
+          const url = new URL(req.url ?? '', `http://${req.headers.host}`);
           if (req.method === 'POST') {
-            const host = req.headers.host ?? 'http://unknown';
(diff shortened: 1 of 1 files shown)


mxschmitt (approved):
test would be lovely!

src/program.ts
```diff
@@ -54,9 +54,9 @@ program
       if (options.port) {
         const sessions = new Map<string, SSEServerTransport>();
         const httpServer = http.createServer(async (req, res) => {
+          const url = new URL(req.url ?? '', `http://${req.headers.host}`);
```
mxschmitt, line 57:
```suggestion
          const url = new URL(`http://${req.headers.host ?? 'localhost'}${req.url || ''}`;);
```
what about this?

Skn0tt: this should be exactly the same, because `req.url` is always absolute.

Skn0tt:
> test would be lovely!

It's hard to test, because apparently our current client doesn't pass the `Host` header. Let's live without a test on this.

PR #71, opened 2025-03-28
chore: stitch together iframes into one tree

Skn0tt:
Based on https://github.com/microsoft/playwright/pull/35396, we can now stitch the iframes into one tree.

I also changed from `.locator('html').ariaSnapshot()` to `.locator('body').ariaSnapshot()`, so we get a more compact tree.
Here's how it looks with `'html'`:

```yaml
- document [ref=s1e2]:
  - heading "Hello" [level=1] [ref=s1e4]
  - iframe [ref=s1e5]:
    - document [ref=f1s1e2]:
      - button "World" [ref=f1s1e4]
```

And here's with `'body'`:

```yaml
- heading "Hello" [level=1] [ref=s1e3]
- iframe [ref=s1e4]:
  - button "World" [ref=f1s1e3]
```

If I was an LLM, i'd definitely prefer the second!

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -118,53 +118,34 @@ export class Context {
   }
 
   private async _allFramesSnapshot(frame: playwright.Page | playwright.FrameLocator): Promise<yaml.Document> {
+    const frameIndex = this._lastSnapshotFrames.push(frame) - 1;
     const snapshotString = await frame.locator('body').ariaSnapshot({ ref: true });
     const snapshot = yaml.parseDocument(snapshotString);
 
-    const skip = new Set<yaml.Node>();
-    const locators = new Map<yaml.Node | yaml.Document | yaml.Pair, playwright.FrameLocator>();
-    await yaml.visitAsync(snapshot, {
-      Scalar: async (key, node, path) => {
-        if (skip.has(node))
-          return;
-
-        if (typeof node.value !== 'string')
-          return;
-
-        let parentFrame: playwright.Page | playwright.FrameLocator = frame;
-        for (const node of path.toReversed()) {
-          if (locators.has(node)) {
-            parentFrame = locators.get(node)!;
-            break;
+    const visit = async (node: any): Promise<unknown> => {
+      if (yaml.isPair(node)) {
+        node.key = await visit(node.key);
+        node.value = await visit(node.value);
+      } else if (yaml.isSeq(node) || yaml.isMap(node)) {
+        node.items = await Promise.all(node.items.map(visit));
+      } else if (yaml.isScalar(node)) {
+        if (typeof node.value === 'string') {
+          const value = node.value;
+          if (frameIndex > 0)
+            node.value = value.replace('[ref=', `[ref=f${frameIndex}`);
+          if (value.startsWith('iframe ')) {
+            const ref = value.match(/\[ref=(.*)\]/)?.[1];
+            if (ref) {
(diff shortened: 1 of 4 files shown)


src/context.ts
```diff
@@ -111,40 +111,39 @@ export class Context {
...
-    return snapshots.join('\n');
+    this._lastSnapshotFrames = [];
+    return await this._allFramesSnapshot(this.existingPage());
+  }
+
+  private async _allFramesSnapshot(frame: playwright.Page | playwright.FrameLocator): Promise<string> {
+    const frameIndex = this._lastSnapshotFrames.push(frame) - 1;
+    const snapshot = await frame.locator('body').ariaSnapshot({ ref: true });
+    const result = await Promise.all(snapshot.split('\n').map(async line => {
+      const scopedLine = frameIndex > 0 ? line.replace('[ref=', `[ref=f${frameIndex}`) : line;
```
pavelfeldman, line 122: This would probably be a good time to parse YAML...

src/context.ts
```diff
@@ -111,40 +112,74 @@ export class Context {
...
+    const yaml = await this._allFramesSnapshot(this.existingPage());
+    return yaml.toString();
+  }
+
+  private async _allFramesSnapshot(frame: playwright.Page | playwright.FrameLocator): Promise<yaml.Document> {
+    const snapshotString = await frame.locator('body').ariaSnapshot({ ref: true });
+    const snapshot = yaml.parseDocument(snapshotString);
+
+    const skip = new Set<yaml.Node>();
+    const locators = new Map<yaml.Node | yaml.Document | yaml.Pair, playwright.FrameLocator>();
```
pavelfeldman, line 125: Looks vibe-coded to me!

Skn0tt: nope, hand-rolled!

Skn0tt:
> Let's make sure we refactor this code for readability / review-ability. I could not review it for correctness. I'm missing stages, variable naming and recursion.

I spent an hour making it readable, this is the best I could come up with. The unnerving thing is that the `.visit` implementation traverses *into* replaced nodes, which means we don't need recursion, but it also makes it hard to understand. Let me try again.

Skn0tt:
Writing my own `visit` made it so much easier. Take another look, it should now be a bog standard traversal.

src/context.ts
```diff
@@ -111,40 +112,74 @@ export class Context {
...
+    await yaml.visitAsync(snapshot, {
```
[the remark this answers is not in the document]
Skn0tt: because taking the snapshot is async

src/context.ts
```diff
@@ -125,40 +126,57 @@ export class Context {
...
+        node.items = await Promise.all(node.items.map(visit));
+      } else if (yaml.isScalar(node)) {
+        if (typeof node.value === 'string') {
+          const value = node.value;
+          if (frameIndex > 0)
+            node.value = value.replace('[ref=', `[ref=f${frameIndex}`);
+          if (value.startsWith('iframe ')) {
+            const ref = value.match(/\[ref=(.*)\]/)?.[1];
+            if (ref) {
+              const childSnapshot = await this._allFramesSnapshot(frame.frameLocator(`aria-ref=${ref}`));
```
pavelfeldman, line 155: I'd i'd silent-catch errors in nested frames just in case

pavelfeldman (approved):

PR #74, opened 2025-03-28
fix: improve description for browser navigation tool

mxschmitt:
Previously things like:

> go to abc.com and let me know whats on there

were not working. This makes it more obvious for the LLM to understand what this tool does.


pavelfeldman:
Screenshot 2025-03-28 at 8 18 09 AM

pavelfeldman:
3.5 does not work, but the problem is not that it could not convert "go" to "navigate". The model is lacking a bit. I'd prefer to keep descriptions precise and narrow to allow for better MCP scale.

Screenshot 2025-03-28 at 11 44 02 AM

PR #76, opened 2025-03-28
chore: isolate SSE client browsers

pavelfeldman:


src/program.ts
```diff
@@ -137,3 +84,64 @@ async function userDataDir() {
...
+
+      await transport.handlePostMessage(req, res);
+      return;
+    } else if (req.method === 'GET') {
+      const transport = new SSEServerTransport('/sse', res);
+      sessions.set(transport.sessionId, transport);
+      const server = await serverList.create();
+      res.on('close', () => {
+        sessions.delete(transport.sessionId);
+        serverList.close(server);
```
dgozman, line 114: This method is async, perhaps `void` it? Maybe enable an eslint rule?

dgozman (approved):

PR #91, opened 2025-03-30
Feat: batch processing

m2rads:
Add a new feature to process tools in a batch call
Merge main from playwright mcp


m2rads:
sorry opened this by mistake

Issue #5, opened 2025-03-22
Feature request: Add viewport size parameters to browser_navigate command


cocosar:
### Description
I'd like to suggest adding viewport size configuration parameters to the `browser_navigate` command. Currently, there's no direct way to specify the browser viewport dimensions when navigating to a URL. Adding this capability would enhance testing flexibility, especially for responsive designs.

### Use Cases
- Testing responsive designs across different screen sizes (mobile, tablet, desktop)
- Ensuring consistent screenshots/snapshots across environments
- Creating more realistic examples for documentation
- Simplifying test workflows by setting viewport dimensions at navigation time

### Proposed Implementation
Add optional parameters to the `browser_navigate` function:
- `width`: The viewport width in pixels (default could remain as current implementation)
- `height`: The viewport height in pixels (default could remain as current implementation)

Example usage:
```javascript
{
  `url`: `https://example.com`,
  `width`: `1920`,
  `height`: `800`
}

- Page URL: https://example.com/
- Page Title: Example Domain
- Page Snapshot
```yaml
- document [ref=s1e2]:
  - heading "Example Domain" [level=1] [ref=s1e5]
  - paragraph [ref=s1e6]: This domain is for use in illustrative examples in documents. You may use this domain in literature without prior coordination or asking for permission.
  - paragraph [ref=s1e7]:
    - link "More information..." [ref=s1e8]:
      - /url: https://www.iana.org/domains/example
```

### Benefits
1. **Flexibility**: Allows testing across different device form factors with a single command
2. **Efficiency**: Reduces the need for separate viewport adjustment commands
3. **User Experience**: More intuitive for demonstrating responsive behaviors
4. **Consistency**: Provides standardized viewport control across different sessions

### Implementation Complexity
This feature should be relatively straightforward to implement as Playwright already supports viewport size manipulation through its built-in APIs (page.setViewportSize() method).

I'd be happy to provide more details or collaborate on implementing this feature if there's interest.

pavelfeldman:
I'll be supportive of the browser_resize command to keep it similar to what end user would do.

PR #92, opened 2025-03-31
feat: browser_resize

Skn0tt:
Closes https://github.com/microsoft/playwright-mcp/issues/5


src/tools/common.ts
```diff
@@ -174,3 +174,28 @@ export const chooseFile: ToolFactory = snapshot => ({
     }, snapshot);
   },
 });
+
+const resizeSchema = z.object({
```
pavelfeldman, line 178: Let them specify pixels? We should be careful about exposing it as it is niche and unclear how it is going to be used. Original request wanted it for responsive design where 800 vs 1024 vs 1280 could result in different media query matches...

pavelfeldman (dismissed):

src/tools/common.ts
Skn0tt: done in https://github.com/microsoft/playwright-mcp/pull/92/commits/5699c1703e51a5739f9b826e301a2c49e83eb00c

pavelfeldman:
Sorry you need a rebase

Skn0tt:
Done!

pavelfeldman (approved):

tests/basic.spec.ts
```diff
@@ -233,3 +233,22 @@ test('browser_type (slowly)', async ({ client }) => {
...
+
+  const response = await client.callTool({
+    name: 'browser_resize',
+    arguments: {
+      width: 390,
+      height: 780,
+    },
+  });
+  expect(response).toContainTextContent('Resized browser window');
+  expect(response).toContainTextContent('Window size: 390x780');
```
yury-s, line 253: This test is flaky: https://github.com/microsoft/playwright-mcp/actions/runs/14458528178/job/40546693304?pr=174, I can also reproduce it locally with `npm run test -- basic.spec.ts:237 --project=chromium --repeat-each 100 -x`

PR #98, opened 2025-03-31
Add functionality to save screenshots as image files

Nazruden:
This PR enhances the screenshot tool to store images as files instead of only returning base64 data. The changes:
* Add filesystem storage for screenshots in a dedicated directory
* Return file path information for screenshots
* Maintain backward compatibility with existing base64 functionality
* Update related documentation and comments
* Implement consistent naming convention for screenshot files

These improvements enable better integration with image analysis tools and simplify workflow when visual content needs to be persisted.


Nazruden:

pavelfeldman:
I've seen other extensions store taken screenshot for the session unconditionally and then exposing them as mcp resources. Even if client struggles with saving tool results, it should be comfortable saving mcp resources.

Nazruden:
The main incentive here was to allow clients like Cursor which don't handle base64 image returns by tools to still be able to get the screenshots.
That's a recurring issue discussed on Cursor's forum these days, I therefore made that change and thought it could be helpful for others.

src/tools/utils.ts
```diff
@@ -69,6 +71,29 @@ export async function waitForCompletion<R>(page: playwright.Page, callback: () =
...
+  // Ensure directory exists
+  await fs.promises.mkdir(directory, { recursive: true });
+
+  // Generate unique filename with timestamp
+  const timestamp = new Date()
+      .toISOString()
+      .replace(/:/g, '-')
+      .replace(/\./g, '-');
+  const filename = `screenshot-${timestamp}.${type}`;
+  const filepath = path.join(directory, sanitizeForFilePath(filename));
```
master96, line 88: This bit actually removes `.` which is part of fileName above.

You could do reverse

```javascript
const filename = sanitizeForFilePath(`screenshot-${timestamp}`);
const filepath = path.join(directory, `${filename}.${type}`);
```

govindrai:
@Nazruden is it possible for you to add in full page screenshotting in this PR? 

```await page.screenshot({ path: 'screenshot.png', fullPage: true });```
[Link](https://playwright.dev/docs/screenshots#full-page-screenshots) to Screenshot API

pavelfeldman:
Screenshots are now stored as files

PR #100, opened 2025-03-31
chore: merge browser and channel settings

pavelfeldman:


src/program.ts
```diff
@@ -35,21 +35,33 @@ const packageJSON = require('../package.json');
...
-    .option('--channel <channel>', 'Channel to use for browser, possible values: chrome, msedge, chromium. Default: chrome')
-    .option('--executable-path <path>', 'Path to the browser executable.')
     .action(async options => {
+      let browserName: 'chromium' | 'firefox' | 'webkit' = 'chromium';
+      if (['chromium', 'firefox', 'webkit'].includes(options.browser))
+        browserName = options.browser;
+      let channel: 'chrome' | 'msedge' | undefined;
+      if (options.browser === 'chrome' || options.browser === 'msedge')
+        channel = options.browser;
+      if (browserName === 'chromium' && !channel)
```
yury-s, line 52: This means 'chromium' always becomes 'chrome', is that what we want?

src/program.ts
```diff
@@ -35,21 +35,33 @@ const packageJSON = require('../package.json');
...
     .option('--headless', 'Run browser in headless mode, headed by default')
+    .option('--port <port>', 'Port to listen on for SSE transport.')
     .option('--user-data-dir <path>', 'Path to the user data directory')
     .option('--vision', 'Run server that uses screenshots (Aria snapshots are used by default)')
-    .option('--port <port>', 'Port to listen on for SSE transport.')
-    .option('--cdp-endpoint <endpoint>', 'CDP endpoint to connect to.')
-    .option('--channel <channel>', 'Channel to use for browser, possible values: chrome, msedge, chromium. Default: chrome')
-    .option('--executable-path <path>', 'Path to the browser executable.')
     .action(async options => {
+      let browserName: 'chromium' | 'firefox' | 'webkit' = 'chromium';
```
dgozman, line 46: I think it's much easier to understand and maintain a big switch over `options.browser` that assigns both `browserName` and `channel` for each value.

src/program.ts
pavelfeldman: I'll add a big switch

yury-s (approved):

PR #113, opened 2025-04-01
chore: update node to 22 in github actions

charpeni:
Existing GitHub Actions are currently using Node 18, but this specific version will reach its end of life at the end of the month. I figured it would make sense to upgrade to the LTS version (22).


charpeni:

pavelfeldman:
We'd like to keep it running on the minimum supported Node version.

PR #121, opened 2025-04-02
feat: add slowly option for typing one character at a time

yury-s:
With current description it's hard to talk LLM into using `browser_press_key` instead of `browser_type`. This makes the following prompt work:

```
Goto https://app.utrsports.net/search?type=players"
Focus in the 'Search' box
Slowly type 'John'.
Wait for 'John' in the drop down list.
```

### DIFF
--- a/src/tools/common.ts
+++ b/src/tools/common.ts
@@ -113,7 +113,7 @@ const pressKeySchema = z.object({
 export const pressKey: (captureSnapshot: boolean) => Tool = captureSnapshot => ({
   schema: {
     name: 'browser_press_key',
-    description: 'Press a key on the keyboard',
+    description: 'Press a key on the keyboard.',
     inputSchema: zodToJsonSchema(pressKeySchema),
   },
   handle: async (context, params) => {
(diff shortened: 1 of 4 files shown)


src/tools/snapshot.ts
```diff
@@ -102,6 +102,7 @@ export const hover: Tool = {
 const typeSchema = elementSchema.extend({
   text: z.string().describe('Text to type into the element'),
   submit: z.boolean().describe('Whether to submit entered text (press Enter after)'),
+  slowly: z.boolean().describe('Wether to type one character at a time. Useful for triggering key handlers in the page. By default entire text is filled in at once.'),
```
pavelfeldman, line 105: This should be optional

src/tools/snapshot.ts
```diff
@@ -124,19 +128,19 @@ export const type: Tool = {
   },
 };
 
-const selectOptionSchema = elementSchema.extend({
+const selectSchema = elementSchema.extend({
```
pavelfeldman, line 131: why the change?

src/tools/common.ts
```diff
@@ -113,7 +113,7 @@ const pressKeySchema = z.object({
 export const pressKey: (captureSnapshot: boolean) => Tool = captureSnapshot => ({
   schema: {
     name: 'browser_press_key',
-    description: 'Press a key on the keyboard',
+    description: 'Press a key on the keyboard.',
```
pavelfeldman, line 116: why the change

pavelfeldman (approved):

PR #128, opened 2025-04-03
Feat : browser_refresh

myselfsaurabh:
# feat: add browser_refresh tool for page reloading

## Overview

Add a new `browser_refresh` tool that allows refreshing the current page. This tool provides a direct way to reload pages without having to re-navigate to the current URL, which is essential for testing state changes and refreshing content.

### Features

The tool:

- Reloads the current page using Playwright's `page.reload()` method.
- Works in both snapshot and vision modes.
- Returns the refreshed page state with appropriate snapshots.
- Maintains the same pattern as other navigation tools.

## Changes

- Add refresh tool implementation in `src/tools/common.ts`
- Register the tool in both snapshot and vision modes in `src/index.ts`
- Add corresponding test case in `tests/basic.spec.ts`
- Update documentation in `README.md` for both modes

## Use Case

This tool addresses a common testing need for verifying state changes in web applications:

When testing applications with asynchronous operations or background processing, QA engineers need to verify that UI state updates correctly after these operations complete. The refresh tool provides a simple way to reload the current page to see the latest state without losing context.

For example, in a scenario where a user adds an item with an initial "Pending" status that should change to "New" after backend processing:

1. Add the item (shows as "Pending").
2. Wait for backend processing to complete.
3. Refresh the page to see the updated status ("New").

Currently, MCP is using `browser_go_back` and `browser_go_forward` for this purpose, which can cause items to be removed from view or reset the application state. A dedicated refresh tool allows testers to simply reload the current page while maintaining context, providing a more reliable way to verify state changes.

## Applicable Scenarios

This tool can be used in various scenarios:

- Verifying form submission results.
- Checking status updates after background processing.
- Testing real-time data updates.
- Confirming changes made by other users or systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ The Playwright MCP server supports the following command-line options:
   - Chrome channels: `chrome-beta`, `chrome-canary`, `chrome-dev`
   - Edge channels: `msedge-beta`, `msedge-canary`, `msedge-dev`
   - Default: `chrome`
+- `--caps <caps>`: Comma-separated list of capabilities to enable, possible values: tabs, pdf, history, wait, files, install. Default is all.
 - `--cdp-endpoint <endpoint>`: CDP endpoint to connect to
 - `--executable-path <path>`: Path to the browser executable
 - `--headless`: Run browser in headless mode (headed by default)
@@ -299,7 +300,7 @@ server.connect(transport);
 
 ### Files and Media
 
-- **browser_choose_file**
+- **browser_file_upload**
   - Description: Choose one or multiple files to upload
   - Parameters:
     - `paths` (array): The absolute paths to the files to upload. Can be a single file or multiple files.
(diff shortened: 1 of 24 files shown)


pavelfeldman:
This server is for general purpose browsing, it is not specialized for testing. Refreshing the page is a rare action during day-to-day browsing. Furthermore, llm can bias towards refresh as a solution to all problems during browsing.

pavelfeldman:
I was thinking of configuring which tools are exposed manually. That way we could have a lot of niche tools that are toggled on and off by the user. Some clients (VS Code) already allow turning tools on and off manually, but it is not yet universal across all of them.

myselfsaurabh:
I agree that configurability could be beneficial for some users. If there’s interest in adding this feature, I’m happy to explore how we could make tools like this toggleable in the future. Do you have any suggestions on how we could implement this across different clients?

PR #133, opened 2025-04-04
Update vscode install for stable

digitarald:
Just released: https://code.visualstudio.com/updates/v1_99

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 A Model Context Protocol (MCP) server that provides browser automation capabilities using [Playwright](https://playwright.dev). This server enables LLMs to interact with web pages through structured accessibility snapshots, bypassing the need for screenshots or visually-tuned models.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,33 +33,30 @@ A Model Context Protocol (MCP) server that provides browser automation capabilit
...
-// Github markdown does not allow linking to `vscode:` directly, so you can use our redirect:
-const urlForGithub = `https://insiders.vscode.dev/redirect?url=${encodeURIComponent(urlForWebsites)}`;
--->
+For manual installation, add the following JSON block to your User Settings (JSON) file in VS Code. You can do this by pressing `Ctrl + Shift + P` and typing `Preferences: Open User Settings (JSON)`.
 
-[<img alt="Install in VS Code Insiders" src="https://img.shields.io/badge/VS_Code_Insiders-VS_Code_Insiders?style=flat-square&label=Install%20Server&color=24bfa5">](https://insiders.vscode.dev/redirect?url=vscode-insiders%3Amcp%2Finstall%3F%257B%2522name%2522%253A%2522playwright%2522%252C%2522command%2522%253A%2522npx%2522%252C%2522args%2522%253A%255B%2522-y%2522%252C%2522%2540playwright%252Fmcp%2540latest%2522%255D%257D)
+Optionally, you can add it to a file called `.vscode/mcp.json` in your workspace. This will allow you to share the configuration with others.
 
-Alternatively, you can install the Playwright MCP server using the VS Code CLI:
+> Note that the `mcp` key is not needed in the `.vscode/mcp.json` file.
```
pavelfeldman, line 44: That's very unfortunate as I used to keep my config files in sync across the ecosystem. Now everyone except for VS Code is aligned?

README.md
```diff
@@ -31,33 +33,30 @@ A Model Context Protocol (MCP) server that provides browser automation capabilit
...
-# For VS Code
-code --add-mcp '{"name":"playwright","command":"npx","args":["@playwright/mcp@latest"]}'
-```
-
-```bash
-# For VS Code Insiders
-code-insiders --add-mcp '{"name":"playwright","command":"npx","args":["@playwright/mcp@latest"]}'
+```json
+{
+  "mcp": {
```
pavelfeldman, line 48: Even though it is not required, you pass it here? If I can have it in here, let's remove the note above as noise!

README.md
digitarald: We can keep the original codeblock, but many readme's for MCP have instructions per editor already as the install experience differs based on files and setting UI

pavelfeldman: I'm not sure I understand, is mcp key valid or not? If it is valid, I'd remove this note.

burkeholland: @digitarald we could remove the block for `mcp.json` and just go with User Settings for simplicity.

PR #151, opened 2025-04-09
Restore `package-lock.json` module hashes

cameronfyfe:
- Adds integrity hashes that were missing for 5 npm packages in `package-lock.json`

<hr />

Is there a reason hashes for some of these dependencies are missing from `package-lock.json`?

Right now these omissions prevent me from packaging a nix derivation for this mcp server directly off this repo (https://github.com/cameronfyfe/nix-mcp-servers/blob/main/pkgs/servers/mcp-server-playwright/default.nix#L17) and I was wondering if they might just be missing due to a bad merge or something odd like that at some point. `npm install` under normal use doesn't seem to care if the hashes are missing and installs the packages anyway, but it's a blocker for hermetic build systems like nix.

If this is intentional for some reason I'm not familiar with feel free to ignore and close.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -377,6 +377,8 @@
     },
     "node_modules/@types/node": {
       "version": "22.13.10",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-22.13.10.tgz",
+      "integrity": "sha512-I6LPUvlRH+O6VRUqYOcMudhaIdUVWfsjnZavnsraHvpBwaEyMN29ry+0UVJhImYL16xsscu0aske3yA+uPOWfw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


Skn0tt:
Thanks for the fix! I don't think there's a good reason, no. I'm a little wary of supply chain attacks so let me double-check if i'm getting the same hashes locally ...

cameronfyfe:
> Thanks for the fix! I don't think there's a good reason, no. I'm a little wary of supply chain attacks so let me double-check if i'm getting the same hashes locally ...

Maybe there's a better way but my method for generating those without updating the package versions was
1. set exact versions for those packages in `package.json` to match the version numbers in `package-lock.json`
2. delete the entries for those packages in `package-lock.json`
3. run `npm install`
4. restore `package.json` to what it was before

Skn0tt (approved):
Verified correct. I used https://github.com/grant-dennison/npm-package-lock-add-resolved, also worked great!

PR #159, opened 2025-04-10
Add scrolltoend feature to enable getting a full page screenshot when we have hidden objects 

murtihash:


Skn0tt (changes requested):
Hi! Could you open an issue with the problem you're trying to solve for, so we can discuss? If we come to an agreement in that, i'm happy to review the PR.

pavelfeldman:
Closing as stale

PR #160, opened 2025-04-11
feat: Add support for additional arguments to pass to the browser

kadoshita:
Hello.

I would like to propose adding a parameter to configure browser command-line arguments.

This feature would be particularly useful when working with media devices, as it would allow users to specify options such as `--use-fake-device-for-media-stream`
For reference, please see the Playwright documentation on browser launch options:
https://playwright.dev/docs/api/class-browsertype#browser-type-launch-option-args

I look forward to hearing your thoughts!


Skn0tt:
Hi @kadoshita! I don't think we want to add this to the official `@playwright/mcp`, but you're more than welcome to use this from your fork.

kadoshita:
@Skn0tt Thank you for your comments.
I understand your perspective.
I have decided to maintain this change in my forked repository.

Issue #154, opened 2025-04-09
Feature request: add `browser_downloads_list`and `browser_download_get` tools to allow discovery and retrieval of downloaded documents


eschnou:
I'm using playwright-mcp to automate site navigation and download of documents. It works great using the current version, however the file is stored in a non-accessible temporary environment. I would like to be able to configure another download destination to be able to properly retrieve the files.

Is this already supported? If not, what would be the best way to approach this? I'd be happy to contribute a pull request, I would however appreciate some pointers.

eschnou:
Hi @Skn0tt,  it could definitively, however at this stage I'm not even sure the file is even stored on disk, when I try to open it in the browser it refers to a 'blob'. In the attachment, you can see the file is not on disk but in a blob.

So, you are correct that my request is maybe not the proper solution. A better approach could be tools that allow to list the download files and retrieve one, in base64, through MCP. This would also be better suited to cases where playwright is running on a remote instance, docker image, etc.

<img width="1037" alt="Image" src="https://github.com/user-attachments/assets/ec93045e-304b-4626-8c7f-13554d9bbfb1" />

Skn0tt:
Makes sense!

hugofreire:
Hi, is there away to access download files, i cant find them in my machine

eschnou:
@hugofreire No. There is however a PR #162 in the work to provide tools to get a downloaded file over MCP.

hugofreire:
> [@hugofreire](https://github.com/hugofreire) No. There is however a PR [#162](https://github.com/microsoft/playwright-mcp/pull/162) in the work to provide tools to get a downloaded file over MCP.

thanks :)

PR #162, opened 2025-04-11
feat: provide downloads as resources

Skn0tt:
Resolves https://github.com/microsoft/playwright-mcp/issues/154.


pavelfeldman:
Ref https://forum.cursor.com/t/mcp-resources-not-working-in-0-46-9/59960

Skn0tt:
i've added a tool to retrieve a download, and added a list of downloads to the snapshot. Do you think there's value in keeping the resource? I guess there might be clients that use them.

Skn0tt:
bankrupting this in favour of a tools-based one

Issue #163, opened 2025-04-11
Element interactions to return `_generateLocatorString` result


mastrzyz:
We are using Playwright MCP to automate a test scenario and at the end we want to generate the actual test code for it , ie : 
```
  if (toolName === "browser_click") {
    const result = await playwrightMCPConnectorClient.callTool({
      name: toolName,
      arguments: {
        ref: params.elementId?.toString(),
        element: params.intent,
      },
    });

      code.push(`await page.${locatorString}.click()`);

```

mastrzyz:
Or a new method `browser_generate_locator_string`

Parameter being `ref`

Output being -> `_generateLocatorString()`

Skn0tt:
related to https://github.com/microsoft/playwright-mcp/issues/123

Skn0tt:
There's no guide for it. The way I use it, I ask the model to generate a test, and I can see that it uses the code the MCP returned via #181. That alone improves code quality, because it uses nice locators.

sonicviz:
OK, I've been doing this already, directing PW via MCP to perform certain actions, then generate a PW test to see if it can replicate it. 
So you're saying https://github.com/microsoft/playwright-mcp/pull/181 will kick in automatically and implicitly return the best locaters (it can) doing it that way? I don't need to refine the prompt or extra wrangling? 
How do you ensure the latest PW MCP is being used? Stop/Start the server pulls in the latest?
Thanks.

PR #181, opened 2025-04-15
feat: respond with action and generated locator

Skn0tt:
Closes https://github.com/microsoft/playwright-mcp/issues/163

### DIFF
--- a/tests/basic.spec.ts
+++ b/tests/basic.spec.ts
@@ -26,6 +26,8 @@ test('browser_navigate', async ({ client }) => {
   })).toHaveTextContent(`
 Navigated to data:text/html,<html><title>Title</title><body>Hello, world!</body></html>
 
+- Action: await page.goto('data:text/html,<html><title>Title</title><body>Hello, world!</body></html>');
+
 - Page URL: data:text/html,<html><title>Title</title><body>Hello, world!</body></html>
 - Page Title: Title
 - Page Snapshot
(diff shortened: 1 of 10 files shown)


tests/basic.spec.ts
```diff
@@ -77,7 +82,10 @@ test('browser_select_option', async ({ client }) => {
       ref: 's1e3',
       values: ['bar'],
     },
-  })).toHaveTextContent(`Selected option in "Select"
+  })).toHaveTextContent(`
+Selected option in "Select"
+
+- Action: await page.getByRole('combobox').selectOption(['bar']);
```
pavelfeldman, line 88: I did not mean for us to generate the full JS code, I thought we would only generate the locator and LLM would use baked-in knowledge to re-create the action based on the locator. But maybe it is fine. Probably needs a better caption than Action to understand that it already happened, but we can follow-up.

mastrzyz: @pavelfeldman I'm the original requestor but are you saying playwright-mcp would return 
`getByRole('combobox')`

instead of 

`await page.getByRole('combobox').selectOption(['bar'])`

?

 I want to replicate exactly what happened when our LLM sent playwright-mcp a command so we can send it back to the test file we are generating.

My worry is if we just return a locator, there might be eventually some disconnect between what I think happened and what actually happened in playwright-mcp

PR #182, opened 2025-04-15
Add element screenshot action for snapshots

AndrewBarzu:
Pull request to allow taking screenshots of specific elements on the screen rather than the whole screen.

Also, changed the current screenshot taking to instead save the screenshot to a temp location. Doing this because the current implementation of returning an ImageContent result does not work with most MCP agents. This is also following the PDF implementation, which also exports to a file.

For large screenshots, it can also happen that the buffer is too large for agents, so saving to a file is preferred as to not lose the screenshot due to an image that is too big to be output in the chat.

### DIFF
--- a/src/tools/snapshot.ts
+++ b/src/tools/snapshot.ts
@@ -213,20 +213,25 @@ const screenshot: Tool = {
     const options: playwright.PageScreenshotOptions = { type: fileType, quality: fileType === 'png' ? undefined : 50, scale: 'css', path: fileName };
     const isElementScreenshot = validatedParams.element && validatedParams.ref;
     return await context.currentTab().runAndWaitWithSnapshot(async snapshot => {
+      let screenshot: Buffer | undefined;
       const code = [
         `// Screenshot ${isElementScreenshot ? validatedParams.element : 'viewport'}`,
       ];
       if (isElementScreenshot) {
         const locator = snapshot.refLocator(validatedParams.ref!);
         code.push(`await page.${await generateLocator(locator)}.screenshot(${javascript.formatObject(options)});`);
-        await locator.screenshot(options);
+        screenshot = await locator.screenshot(options);
       } else {
         code.push(`await page.screenshot(${javascript.formatObject(options)});`);
-        await tab.page.screenshot(options);
+        screenshot = await tab.page.screenshot(options);
       }
       code.push(`// Screenshot saved as ${fileName}`);
       return {
-        code
+        code,
+        images: [{
+          data: screenshot.toString('base64'),
+          mimeType: fileType === 'png' ? 'image/png' : 'image/jpeg',
+        }]
       };
     }, { captureSnapshot: false });
   }
--- a/src/tools/snapshot.ts
+++ b/src/tools/snapshot.ts
@@ -209,9 +209,15 @@ const screenshot: Tool = {
     } else {
       return await context.currentTab().runAndWaitWithSnapshot(async snapshot => {
         const locator = snapshot.refLocator(validatedParams.ref);
+        const code = [
+          `// Screenshot ${validatedParams.element}`,
+          `await page.${await generateLocator(locator)}.screenshot(${javascript.formatObject(options)});`
+        ];
         await locator.screenshot(options);
-      }, {
-        status: `Saved as "${fileName}"`,
+        code.push(`// Saved as ${fileName}`);
+        return {
+          code
+        };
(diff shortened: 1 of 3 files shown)


AndrewBarzu:

Skn0tt (commented):
nice! the filename change sounds good.

we discussed this offline. this is useful e.g. for approval processes where screenshots of specific parts of a site need to be handed in.

please experiment with turning this into optional parameters on the existing tool.

pavelfeldman:
For the saving as file party, Is Claude happy with it in terms of reading the screenshot from disk and clicking the right location?

AndrewBarzu:
@pavelfeldman  tried in Claude 3.7 and 3.5, and both seem fine with both finding the right locator on the page, and also with clicking it and then showing the screenshot in chat. Claude needs a bit more pushing than GPT 4/4o which straight up give me the path, for Claude I need to specifically ask for the saved path, but they do cooperate.

src/tools/snapshot.ts
```diff
@@ -183,8 +186,9 @@ const selectOption: Tool = {
   },
 };
 
-const screenshotSchema = z.object({
+const screenshotSchema = elementSchema.extend({
   raw: z.boolean().optional().describe('Whether to return without compression (in PNG format). Default is false, which returns a JPEG image.'),
+  wholePage: z.boolean().optional().describe('Whether to take a screenshot of the whole page. Default is true, which takes a screenshot of the viewport.'),
```
pavelfeldman, line 191: This will clash with fullPage which is used for non-viewported capture.

src/tools/snapshot.ts
```diff
@@ -183,8 +186,9 @@ const selectOption: Tool = {
   },
 };
 
-const screenshotSchema = z.object({
+const screenshotSchema = elementSchema.extend({
```
pavelfeldman, line 189: This is confusing as the element ref becomes required. It should be optional, and it should control whether this is an element screenshot or a viewport screenshot.

src/tools/snapshot.ts
```diff
@@ -198,14 +202,34 @@ const screenshot: Tool = {
   handle: async (context, params) => {
     const validatedParams = screenshotSchema.parse(params);
     const tab = context.currentTab();
-    const options: playwright.PageScreenshotOptions = validatedParams.raw ? { type: 'png', scale: 'css' } : { type: 'jpeg', quality: 50, scale: 'css' };
-    const screenshot = await tab.page.screenshot(options);
+    const fileName = path.join(os.tmpdir(), sanitizeForFilePath(`page-${new Date().toISOString()}`)) + `${validatedParams.raw ? '.png' : '.jpeg'}`;
+    const options: playwright.PageScreenshotOptions = validatedParams.raw ? { type: 'png', scale: 'css', path: fileName } : { type: 'jpeg', quality: 50, scale: 'css', path: fileName };
```
pavelfeldman, line 206: go for default quality here (remove param)

src/tools/snapshot.ts
```diff
@@ -198,14 +202,34 @@ const screenshot: Tool = {
...
+    if (validatedParams.wholePage) {
+      await tab.page.screenshot(options);
```
pavelfeldman, line 208: This needs proper code generated. Like

```js
        const code = [
          `// Save screenshot as ${fileName}`,
          `await page.screenshot(${javascript.formatObject(options)});`
        ];
        return { code };
```

Also move into runAndWaitWithSnapshot to emit updated snapshot.

AndrewBarzu:
> I don't think this is the case as Claude can't open the file you saved the screenshot into. The issue with your approach is that previously, I could ask Claude if my page has rounded borders or what color the button is. With your approach it becomes impossible.

I can't do this with the current implementation either, because aria snapshots don't contain css. Screenshot created by returning an ImageResult are not captured in VSCode agents, and for me every agent I try to take a screenshot through always gives me a result of the form "I attempted to take a screenshot, but it seems there was an issue with the screenshot capture. Let me try again with a different approach." This is because the screenshot is simply not returned in these agents, the ImageResult type of return is not supported, hence why I chose to save in a file and return the path as text. This way, someone can choose to add the screenshot to context and ask questions based on that.

src/tools/snapshot.ts
AndrewBarzu: This also had `options` previously. Without the param we can't specify file format + file location to save the screenshot.

PR #200, opened 2025-04-16
feat: add persistent browser context handling

amitdeshmukh:
- Add _persistentContext property to store and reuse browser context
- Cache and reuse existing persistent context if available. This ensures subsequent tool calls do not fail with the error `Failed to create a ProcessSingleton for your profile directory`
- Improve browser context management efficiency


src/context.ts
```diff
@@ -162,10 +162,18 @@ export class Context {
     return { browserContext };
   }
 
+  private _persistentContext: playwright.BrowserContext | null = null;
+
   private async _launchPersistentContext(): Promise<playwright.BrowserContext> {
+    // If we already have a persistent context, return it.
+    if (this._persistentContext) {
```
pavelfeldman, line 169: It should be higher up, in _ensureBrowserContext

pavelfeldman:
I think I fixed it in one of the drive-bys. Thanks for reporting it at the first place.

PR #204, opened 2025-04-16
chore: introduce modal states

pavelfeldman:


src/context.ts
```diff
@@ -33,20 +34,43 @@ type PageOrFrameLocator = playwright.Page | playwright.FrameLocator;
...
+
+  clearModalState(modalState: ModalState) {
+    this._modalStates = this._modalStates.filter(state => state !== modalState);
+  }
+
+  modalStatesMarkdown(): string[] {
+    const result: string[] = ['### Modal state'];
+    for (const state of this._modalStates) {
+      const tool = this.tools.find(tool => tool.clearsModalState === state.type);
+      result.push(`- [${state.description}]: can be handled by the "${tool?.schema.name}" tool`);
```
yury-s, line 69: Why can't we return this once in the tool's description?

yury-s (approved):

src/context.ts
pavelfeldman: We want to repeat it every time it attempts to click while modal state is on.

src/server.ts
```diff
@@ -57,9 +57,21 @@ export function createServerWithTools(options: Options): Server {
...
+    const modalStates = context.modalStates().map(state => state.type);
+    if ((tool.clearsModalState && !modalStates.includes(tool.clearsModalState)) ||
+        (!tool.clearsModalState && modalStates.length)) {
+      const text = [
+        `Tool "${request.params.name}" does not handle the modal state.`,
+        ...context.modalStatesMarkdown(),
+      ].join('\n');
+      return {
+        content: [{ type: 'text', text }],
+        isError: true,
```
Skn0tt, line 69: I'm a little wary that the model perceives this as a broken tool, not as wrong usage, and that it stops using the tool entirely. https://modelcontextprotocol.io/docs/concepts/tools#error-handling-2 mentions something about human intervention, which could be interpreted like that. Let's keep an eye out.

PR #205, opened 2025-04-17
chore: flatten tool calling, prep for timeout handling

pavelfeldman:


src/tools/files.ts
```diff
@@ -25,27 +25,35 @@ const uploadFileSchema = z.object({
...
-    return await tab.runAndWait(async () => {
-      const modalState = context.modalStates().find(state => state.type === 'fileChooser');
-      if (!modalState)
-        throw new Error('No file chooser visible');
+    const modalState = context.modalStates().find(state => state.type === 'fileChooser');
+    if (!modalState)
+      throw new Error('No file chooser visible');
+
+    const code = [
+      `// <internal code to chose files ${validatedParams.paths.join(', ')}`,
```
yury-s, line 42:
```suggestion
      `// internal code to choose files ${validatedParams.paths.join(', ')}`,
```

src/tools/snapshot.ts
```diff
@@ -57,15 +62,20 @@ const click: Tool = {
...
+    const locator = tab.snapshotOrDie().refLocator(validatedParams.ref);
+
+    const code = [
+      `// Click ${validatedParams.element}`,
+      `await page.${await generateLocator(locator)}.click();`
+    ];
+
+    return {
+      code,
+      action: () => locator.click().then(() => ({})),
```
yury-s, line 75: Why do you nee `.then(() => ({})`?

src/tools/pdf.ts
```diff
@@ -21,27 +21,35 @@ import { z } from 'zod';
...
+      `await page.pdf(${javascript.formatObject({ path: fileName })});`,
+    ];
+
     return {
-      content: [{
-        type: 'text',
-        text: `Saved as ${fileName}`,
-      }],
+      code,
+      action: async () => tab.page.pdf({ path: fileName }).then(() => ({})),
```
yury-s, line 50: just handle `undefined` result once at the call site?

yury-s (approved):

src/tools/files.ts
pavelfeldman: We actually wrap internal code with < >

PR #208, opened 2025-04-17
chore: test list tabs

pavelfeldman:


src/tools/pdf.ts
```diff
@@ -47,7 +47,7 @@ const pdf: Tool = {
 
     return {
       code,
-      action: async () => tab.page.pdf({ path: fileName }).then(() => ({})),
+      action: async () => tab.page.pdf({ path: fileName }).then(() => {}),
```
Skn0tt, line 50:
```suggestion
      action: async () => tab.page.pdf({ path: fileName }),
```

src/tools/snapshot.ts
```diff
@@ -181,7 +180,7 @@ const type: Tool = {
 
     return {
       code,
-      action: () => steps.reduce((acc, step) => acc.then(step), Promise.resolve()).then(() => ({})),
+      action: () => steps.reduce((acc, step) => acc.then(step), Promise.resolve()),
```
Skn0tt, line 183: this is the most confusing for loop i've ever seen

Skn0tt (approved):

PR #209, opened 2025-04-17
feat: `browser_file_download`

Skn0tt:
Resolves https://github.com/microsoft/playwright-mcp/issues/154


Skn0tt:
Summary from offline discussion:

- similar to how a browser starts downloading files without user approval, we can do the same here
- for bigger files, we'll need some sort of a "downloading" state. out of scope for now, we expect most users of this to be on fast connections and to be downloading small files
- for detecting downloads in response to a navigation, checks on error message aren't good. we can check if playwright-core can detect this better.

jmmfcoutinho:
Is this PR closed without any merge? Do you have any estimate when the `browser_download` feature will be available?

Skn0tt:
don't worry José, we're continuing work in https://github.com/microsoft/playwright-mcp/pull/310 :D

PR #212, opened 2025-04-17
feat(dialog): handle dialogs

pavelfeldman:


src/tools/utils.ts
```diff
@@ -62,7 +63,7 @@ export async function waitForCompletion<R>(page: playwright.Page, callback: () =
     if (!requests.size && !frameNavigated)
       waitCallback();
     await waitBarrier;
-    await page.evaluate(() => new Promise(f => setTimeout(f, 1000)));
+    await context.waitForTimeout(1000);
```
yury-s, line 66: Why do we need this?

src/context.ts
```diff
@@ -190,6 +197,43 @@ ${code.join('\n')}
     };
   }
 
+  async waitForTimeout(time: number) {
+    if (this._currentTab && !this._javaScriptBlocked())
```
yury-s, line 201: Why not always `await new Promise(f => setTimeout(f, time));` ?

src/context.ts
```diff
@@ -190,6 +197,43 @@ ${code.join('\n')}
...
+      await this._currentTab.page.waitForTimeout(time);
+    else
+      await new Promise(f => setTimeout(f, time));
+  }
+
+  private async _raceAgainstModalDialogs(action: () => Promise<ToolActionResult>): Promise<ToolActionResult> {
+    this._pendingAction = {
```
yury-s, line 208: Check that it's not currently set.

yury-s (approved):

src/tools/utils.ts
pavelfeldman: For "stability"

src/context.ts
pavelfeldman: We are past that, concurrent tool invocation is "undefined behavior" as of today.

src/context.ts
pavelfeldman: To make align with the browser clock.

yury-s (approved):

PR #240, opened 2025-04-22
fix: show custom error for modal state

Skn0tt:
Calling a tool that resolves modal state, when there's no such modal state visible, currently shows this misleading message:

```md
Tool "browser_file_upload" does not handle the modal state.
### Modal state
```

Instead, we should show the error message from the tool implementation.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -49,33 +49,25 @@ export function createServerWithTools(options: Options): Server {
   });
 
   server.setRequestHandler(CallToolRequestSchema, async request => {
+    const errorResult = (...messages: string[]) => ({
+      content: [{ type: 'text', text: messages.join('\n') }],
+      isError: true,
+    });
     const tool = tools.find(tool => tool.schema.name === request.params.name);
-    if (!tool) {
-      return {
-        content: [{ type: 'text', text: `Tool "${request.params.name}" not found` }],
-        isError: true,
-      };
-    }
+    if (!tool)
+      return errorResult(`Tool "${request.params.name}" not found`);
+
 
     const modalStates = context.modalStates().map(state => state.type);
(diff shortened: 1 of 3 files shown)


pavelfeldman:
I'd prefer to stick with the fail-fast principle, i.e. handle errors as early as they can be detected. Calling a tool that clears modal state w/o that modal state is an error. Showing empty list is not intuitive, maybe we can render "No modal states" instead.

Skn0tt:
Makes me think we could also only conditionally make those tools available when there's a modal change. I'll take a separate look at that - how do you like the new error message?

src/server.ts
```diff
@@ -49,34 +49,25 @@ export function createServerWithTools(options: Options): Server {
...
-        `Tool "${request.params.name}" does not handle the modal state.`,
-        ...context.modalStatesMarkdown(),
-      ].join('\n');
-      return {
-        content: [{ type: 'text', text }],
-        isError: true,
-      };
-    }
+    if (tool.clearsModalState && !modalStates.includes(tool.clearsModalState))
+      return errorResult(`The tool "${request.params.name}" can only be used when there is related modal state present.`, ...context.modalStatesMarkdown());
```
pavelfeldman, line 63: This sounds like a very inefficient way of providing the feedback. Should llm trial all tools based on it and find the right one? Or we already list the exact tools it can use?

Skn0tt: `modalStatesMarkdown` should contain the exact tools, yes

Skn0tt: `modalStatesMarkdown` should contain the currently opened modal states and what tools can be used to interact with it, yes

pavelfeldman (approved):

PR #241, opened 2025-04-22
chore: infer tool params

Skn0tt:
Moves the `schema.parse` call to the calling side of the handler, so we don't have to duplicate it everywhere.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -124,7 +124,7 @@ export class Context {
 
   async run(tool: Tool, params: Record<string, unknown> | undefined) {
     // Tab management is done outside of the action() call.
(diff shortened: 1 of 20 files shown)


dgozman (commented):
Code looks good, but bots are unhappy.

Skn0tt:
> Code looks good, but bots are unhappy.

Wonder why I didn't see them locally 🤔 

The tests were off - all MCP tools require passing an argument, even if it's the empty object. We didn't fail before because some of our tools didn't validate their schema. Now that the schema checking is centralized, the tests fail. Fixed it by updating the tests.

dgozman (approved):

PR #243, opened 2025-04-22
feat: support streamable http transport

Skn0tt:
Adds support for the new StreamableHttp transport. I'm not aware of any clients that implement it, but somebody's gotta make the start! Once some clients support it, we can also advertise it in the README.

### DIFF
--- a/src/program.ts
+++ b/src/program.ts
@@ -15,10 +15,12 @@
  */
 
 import http from 'http';
+import crypto from 'node:crypto';
 
 import { program } from 'commander';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
 import { SSEServerTransport } from '@modelcontextprotocol/sdk/server/sse.js';
+import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
 
 
 import { createServer } from './index';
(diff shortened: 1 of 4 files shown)


src/program.ts
```diff
@@ -74,9 +76,45 @@ function setupExitWatchdog(serverList: ServerList) {
 
 program.parse(process.argv);
 
-async function startSSEServer(port: number, serverList: ServerList) {
-  const sessions = new Map<string, SSEServerTransport>();
+async function startHttpServer(port: number, serverList: ServerList) {
+  const sseSessions = new Map<string, SSEServerTransport>();
+  const streamableSessions = new Map<string, StreamableHTTPServerTransport>();
   const httpServer = http.createServer(async (req, res) => {
+    const isStreamableHttp = req.url?.startsWith('/mcp');
```
pavelfeldman, line 83: Please please kick this code into a separate file!

Skn0tt: done, and reorganised it

src/transport.ts
```diff
@@ -0,0 +1,128 @@
...
+import { SSEServerTransport } from '@modelcontextprotocol/sdk/server/sse.js';
+import { StreamableHTTPServerTransport } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
+
+export async function startStdioTransport(serverList: ServerList) {
+  const server = await serverList.create();
+  await server.connect(new StdioServerTransport());
+}
+
+export async function startHttpTransport(port: number, serverList: ServerList) {
+  const sseSessions = new Map<string, SSEServerTransport>();
```
pavelfeldman, line 32: Is there a reason we keep these in the same namespace / closure? To me those are two different servers.

pavelfeldman (approved):

mxschmitt (approved):

PR #251, opened 2025-04-23
feat: add request and response body on network tool

bun913:
## Background

Closes https://github.com/microsoft/playwright-mcp/issues/271

## Details

- I modified `network.ts` to enable it to return `request` and `response` body information
    - And it returns only `xhr` or `fetch` request, because the context window is not so big, so big CSS and fonts are a waste of space
- I added the test on `network.spec.ts`

### DIFF
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -47,12 +47,7 @@ const requests = defineTool({
 
 // Only include request/response body for XHR/fetch/json requests
 function shouldIncludeBody(request: playwright.Request): boolean {
-  const resourceType = request.resourceType();
-  if (resourceType === 'xhr' || resourceType === 'fetch')
-    return true;
   const headers = request.headers();
-  if (headers['x-requested-with']?.toLowerCase() === 'xmlhttprequest')
-    return true;
   return request.isNavigationRequest() === false &&
          (headers.accept?.includes('application/json') ||
           headers['content-type']?.includes('application/json'));
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -36,7 +36,7 @@ const requests = defineTool({
       code: [`// <internal code to list network requests>`],
       action: async () => {
         return {
-          content: [{ type: 'text', text: log.join('\n') }]
+          content: [{ type: 'text', text: log }]
         };
       },
       captureSnapshot: false,
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -24,7 +24,7 @@ const requests = defineTool({
 
   schema: {
     name: 'browser_network_requests',
-    description: 'Returns all network requests since loading the page, if the request is a xhr or fetch, it will return the request body and response body',
+    description: 'Returns all network requests since loading the page',
     inputSchema: z.object({}),
   },
 
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -33,7 +33,7 @@ const requests = defineTool({
     const allRequests = [...requests.entries()];
     const log = await Promise.all(allRequests.map(async ([request, response]) => await renderRequest(request, response)));
     return {
-      code: ['// <internal code to list all network requests>'],
+      code: [`// <internal code to list all network requests>`],
       action: async () => {
         return {
           content: [{ type: 'text', text: log }]
--- a/src/tools/network.ts
+++ b/src/tools/network.ts
@@ -24,15 +24,19 @@ const requests = defineTool({
 
   schema: {
     name: 'browser_network_requests',
-    description: 'Returns all network requests since loading the page',
+    description: 'Returns XHR/Fetch API network requests since loading the page',
     inputSchema: z.object({}),
   },
 
   handle: async context => {
     const requests = context.currentTabOrDie().requests();
-    const log = await Promise.all([...requests.entries()].map(async ([request, response]) => renderRequest(request, response)));
+
+    // Filter XHR requests
+    const xhrRequests = [...requests.entries()].filter(([request]) => isXhrRequest(request));
+
+    const log = await Promise.all(xhrRequests.map(async ([request, response]) => renderRequest(request, response)));
     return {
-      code: [`// <internal code to list network requests>`],
+      code: [`// <internal code to list XHR network requests>`],
       action: async () => {
         return {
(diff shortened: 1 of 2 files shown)


bun913:

Skn0tt (commented):
I like adding the request/response bodies, but I don't think we should limit the tool result to only XHR/fetch requests. Could you trim down the PR to just that? Maybe you could conditionally only include the bodies for XHR/fetch requests.

bun913:
@Skn0tt 
Thank you for your advice. As you mentioned, it is better to include all requests.
So, I changed it like below.

- XHR or fetch
    - Include Request and Response Body
- except them
    - Include the `[METHOD] URL` only

src/tools/network.ts
```diff
@@ -44,11 +45,37 @@ const requests = defineTool({
...
+function shouldIncludeBody(request: playwright.Request): boolean {
+  const resourceType = request.resourceType();
+  if (resourceType === 'xhr' || resourceType === 'fetch')
+    return true;
+  const headers = request.headers();
+  if (headers['x-requested-with']?.toLowerCase() === 'xmlhttprequest')
+    return true;
+  return request.isNavigationRequest() === false &&
+         (headers.accept?.includes('application/json') ||
+          headers['content-type']?.includes('application/json'));
```
Skn0tt, line 58: let's just check if there's a body, not what its content type is. if the request has a body, that's always interesting.

src/tools/network.ts
```diff
@@ -44,11 +45,37 @@ const requests = defineTool({
...
-function renderRequest(request: playwright.Request, response: playwright.Response | null) {
+// Only include request/response body for XHR/fetch/json requests
+function shouldIncludeBody(request: playwright.Request): boolean {
+  const resourceType = request.resourceType();
+  if (resourceType === 'xhr' || resourceType === 'fetch')
+    return true;
+  const headers = request.headers();
+  if (headers['x-requested-with']?.toLowerCase() === 'xmlhttprequest')
+    return true;
+  return request.isNavigationRequest() === false &&
```
Skn0tt, line 56: a form submission could be a navigation request, but still has a super interesting body. let's not check for it being a nav request.

src/tools/network.ts
```diff
@@ -44,11 +45,37 @@ const requests = defineTool({
...
-  if (response)
-    result.push(`=> [${response.status()}] ${response.statusText()}`);
+
+  // Add request body for XHR/fetch/json only
+  if (shouldIncludeBody(request)) {
+    const postData = request.postData();
+    if (postData)
+      result.push(`Request Body: ${postData}`);
+    if (response) {
+      const body = await response.body();
```
Skn0tt, line 71: use `response.text()`

tests/network.spec.ts
```diff
@@ -42,8 +47,150 @@ test('browser_network_requests', async ({ client, server }) => {
...
+  })).toHaveTextContent(`Request Body: {"data":"test payload"}`);
+
+  // Check if response body is displayed
+  expect.poll(() => client.callTool({
+    name: 'browser_network_requests',
+    arguments: {},
+  })).toHaveTextContent(`Response Body: {"name":"John Doe"}`);
+});
+
+test('browser_network_requests_get_xhr', async ({ client, server }) => {
```
Skn0tt, line 68:
```suggestion
test('browser_network_requests_with_bodies', async ({ client, server }) => {
```

src/tools/network.ts
```diff
@@ -24,15 +24,16 @@ const requests = defineTool({
...
   },
 
   handle: async context => {
     const requests = context.currentTabOrDie().requests();
-    const log = [...requests.entries()].map(([request, response]) => renderRequest(request, response)).join('\n');
+    const allRequests = [...requests.entries()];
+    const log = await Promise.all(allRequests.map(async ([request, response]) => renderRequest(request, response)));
     return {
-      code: [`// <internal code to list network requests>`],
+      code: ['// <internal code to list all network requests>'],
```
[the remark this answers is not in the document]
bun913: https://github.com/microsoft/playwright-mcp/pull/251/commits/9341e64c587eb93a25e607abb9a396c07bf5c8df

I revert as above

src/tools/network.ts
```diff
@@ -24,15 +24,16 @@ const requests = defineTool({
 
   schema: {
     name: 'browser_network_requests',
-    description: 'Returns all network requests since loading the page',
+    description: 'Returns all network requests since loading the page, if the request is a xhr or fetch, it will return the request body and response body',
```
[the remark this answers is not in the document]
bun913: I revert this line

src/tools/network.ts
bun913: I fixed the conditions like the below

https://github.com/microsoft/playwright-mcp/pull/251/commits/996df4f125f3aa673563b6401588bc2490a306ba

src/tools/network.ts
bun913: I fixed the conditions same as the above comment.

- Is application/json request?
- And if there is body, the tool includes it in the result

src/tools/network.ts
bun913: Fixed

https://github.com/microsoft/playwright-mcp/pull/251/commits/996df4f125f3aa673563b6401588bc2490a306ba#diff-5572ae9b262f5d38deefdfe1e74289040344421e86c205e68902ce0d34e53becR63

bun913:
@Skn0tt 

Sorry for bothering you.
I fixed and simplified the overall code as you mentioned.
Please check them at your earliest convenience.

bun913:
I fixed some lint errors and ran `npm run lint` to check them.

praveen2399:
I am looking for this feature. Can this be available in next release ?

bun913:
@praveen2399 
Thank you. I'm also finding that the response body is beneficial when the request is in JSON format.

@Skn0tt 

Could you please re-review at your earliest convenience?

pavelfeldman:
> Thank you. I'm also finding that the response body is beneficial when the request is in JSON format.

That's hardly an excuse to flood the chat with the bodies. I'll close this for now as non-practical. I think we should log network into files (har) and expose them as resources for the next gen clients though.

PR #268, opened 2025-04-25
chore: don't fail when browser is already installed

Skn0tt:
Some models like calling `browser_install` just to be sure. For users of channels like chrome or msedge, this call throws an error; making the model think that it can't use a browser at all!

### DIFF
--- a/src/tools/install.ts
+++ b/src/tools/install.ts
@@ -34,19 +34,22 @@ const install = defineTool({
     const child = fork(cli, ['install', channel], {
       stdio: 'pipe',
     });
-    const output: string[] = [];
-    child.stdout?.on('data', data => output.push(data.toString()));
-    child.stderr?.on('data', data => output.push(data.toString()));
-    await new Promise<void>((resolve, reject) => {
+    const chunks: string[] = [];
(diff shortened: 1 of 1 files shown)


src/tools/install.ts
```diff
@@ -34,19 +34,22 @@ const install = defineTool({
...
+    const chunks: string[] = [];
+    child.stdout?.on('data', data => chunks.push(data.toString()));
+    child.stderr?.on('data', data => chunks.push(data.toString()));
+    const code = await new Promise<string>((resolve, reject) => {
       child.on('close', code => {
+        const output = chunks.join('');
         if (code === 0)
-          resolve();
+          resolve(`// Browser ${channel} installed`);
+        else if (output.includes('already installed'))
```
pavelfeldman, line 45: That's going to break over time, especially if we don't have a test.

Skn0tt: What'd you recommend instead?

Skn0tt:
closing in favour of https://github.com/pulls?q=is%3Aopen+is%3Apr+author%3ASkn0tt+archived%3Afalse+org%3Amicrosoft

PR #273, opened 2025-04-26
Optimize Dockerfile by excluding unnecessary files and using non-root user

KanchiShimono:
## Changes

- Exclude unnecessary source files and development `node_modules` from the final image
- Switch runtime user from `root` to non-root `node` user

## Context

- Improve container security
- Minimize runtime privileges

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN --mount=type=cache,target=/root/.npm,sharing=locked,id=npm-cache \
   # Install `node_modules` including dev dependencies to run playwright commands with npx
   npm ci && \
   # Install system dependencies for playwright
-  npx playwright install-deps chromium-headless-shell && \
+  npx -y playwright install-deps chromium && \
   # Overwrite with only production `node_modules` required for runtime
   npm ci --omit=dev
 
@@ -33,7 +33,7 @@ RUN --mount=type=cache,target=/root/.npm,sharing=locked,id=npm-cache \
   # Install `node_modules` including dev dependencies to run playwright commands with npx
   npm ci && \
   # Install browser binaries
-  npx playwright install --only-shell chromium
+  npx -y playwright install --no-shell chromium
 
 # Copy the rest of the app
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ ARG PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
 # ------------------------------
 # Base
 # ------------------------------
+# Base stage: Contains only the minimal dependencies required for runtime
+# (node_modules and Playwright system dependencies)
 FROM node:22-bookworm-slim AS base
 
 ARG PLAYWRIGHT_BROWSERS_PATH
@@ -38,6 +40,9 @@ RUN npm run build
 # ------------------------------
 # Browser
 # ------------------------------
+# Cache optimization:
+# - Browser is downloaded only when node_modules or Playwright system dependencies change
+# - Cache is reused when only source code changes
 FROM base AS browser
 
 RUN npx -y playwright-core install --no-shell chromium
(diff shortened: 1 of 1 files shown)


Skn0tt:
@cmrigney could you take a look?

Dockerfile
```diff
@@ -1,22 +1,67 @@
...
 
-# Install chromium and its dependencies, but only for headless mode
-RUN npx -y playwright install --with-deps --only-shell chromium
+RUN --mount=type=cache,target=/root/.npm,sharing=locked,id=npm-cache \
+    --mount=type=bind,source=package.json,target=package.json \
+    --mount=type=bind,source=package-lock.json,target=package-lock.json \
+  # Install `node_modules` including dev dependencies to run playwright commands with npx
+  npm ci && \
+  # Install browser binaries
+  npx playwright install --only-shell chromium
```
pavelfeldman, line 36: `--only-shell` is incompatible with the server as we are using new headless (from chromium) on ToT.

cmrigney (approved):
lgtm 👍 
Nice improvements!

Dockerfile
KanchiShimono: @pavelfeldman 
Thank you for your reply!

I’ve updated the browser executable being installed to use the new headless mode.
Since the new headless mode requires specifying the `chromium` channel in `LaunchOptions`, I also made a small change to the source code.

mxschmitt (commented):
I've got a few questions regarding this patch:

- Why do you want to make the username configurable? Why not always run it with `node` user?
- `COPY .` was used, I think we should avoid that due to caching issues
- If we always run it with `node` user we can remove the `PLAYWRIGHT_BROWSERS_PATH` handling - makes it simpler
- I don't see a reason why we can't install the browsers in the runtime docker image, this would make things simpler - doesn't it?
- This means the multi stage build is only for saving NPM dev dependencies

I took a stab and [the following](https://gist.github.com/mxschmitt/7318a392c543660cfe31988cd46699b7) worked for me, do you see any disadvantages in that approach?

cmrigney:
@mxschmitt Regarding this:
> I don't see a reason why we can't install the browsers in the runtime docker image, this would make things simpler - doesn't it?

I didn't check, but do you know the difference in image size for headless chromium versus including all the browsers? If it's a big difference, maybe it would be better to have build targets for each different browser. [Bake](https://docs.docker.com/build/bake/) is useful for that kind of thing.

KanchiShimono:
@mxschmitt 
Thank you for the review.
The `Dockerfile` in the gist looks great!

That said, I still believe there are advantages to using `PLAYWRIGHT_BROWSERS_PATH` and keeping the browser installation in a separate build stage, for the following two reasons:

1. Why use `PLAYWRIGHT_BROWSERS_PATH`:
In some cases, external volumes may be mounted to $HOME for persisting user data directories or screenshots. Since Playwright installs browsers under $HOME by default, these binaries can get wiped out by the volume mount. To avoid this, we explicitly set a custom browser installation path.

2. Why separate the browser installation stage:
In addition to the points mentioned by @cmrigney, installing browsers during the runtime stage means we need to be careful with the placement of the install step to avoid reinstalling browsers on every code change. Separating the installation into its own stage allows us to simply `COPY` the binaries, making the build more cache-friendly and easier to maintain.

### Answers to your questions

> - Why do you want to make the username configurable? Why not always run it with `node` user?
> - `COPY .` was used, I think we should avoid that due to caching issues

These are valid suggestions.

> - If we always run it with node user we can remove the PLAYWRIGHT_BROWSERS_PATH handling – makes it simpler
> - I don’t see a reason why we can’t install the browsers in the runtime Docker image; this would make things simpler – doesn’t it?

I wasn’t aware that we could install browsers using `playwright-core` without needing dev dependencies in `node_modules`. That’s why I originally installed them during the build stage.
That said, for the reasons above, I still believe there’s value in specifying a custom browser path and separating the browser installation into its own stage.

KanchiShimono:
The image size increases by approximately 150MB between the new headless mode (no-shell) and the old headless mode (headless-shell).

Output of `nerdctl images`:
```
REPOSITORY        TAG               SIZE       BLOB SIZE
playwright-mcp    no-shell          1.257GB    414.5MB
playwright-mcp    headless-shell    1.092GB    368.5MB
```

KanchiShimono:
I successfully captured a screenshot using the `Dockerfile` from this PR and the latest main branch (23ce973).
I'm using `@modelcontextprotocol/inspector` as MCP client.

スクリーンショット 2025-05-02 14 25 02

KanchiShimono:
@pavelfeldman @mxschmitt @cmrigney 
Could you please re-review this PR?

As @cmrigney mentioned, the latest container image from the `main` branch is no longer working.

Dockerfile
```diff
@@ -1,22 +1,69 @@
...
-RUN npx -y playwright install --with-deps --only-shell chromium
+RUN --mount=type=cache,target=/root/.npm,sharing=locked,id=npm-cache \
+    --mount=type=bind,source=package.json,target=package.json \
+    --mount=type=bind,source=package-lock.json,target=package-lock.json \
+  npm ci
 
 # Copy the rest of the app
-COPY . .
+COPY *.json *.js *.ts .
+COPY src src/
```
cmrigney, line 35: Not yet stable, but there will be an `--exclude` flag coming so that we could eventually replace these 2 with one COPY command - https://docs.docker.com/reference/dockerfile/#copy---exclude

cmrigney (approved):
Dockerfile lgtm.

I still don't understand why it wouldn't take a screenshot for me, but may be unrelated to this PR.

PR #274, opened 2025-04-27
feat: add proxy server option to set a global proxy server for called browser

adamhj:
This commit introduces `--proxy-server` and `--proxy-bypass` parameter to the CLI, allowing users to config a global proxy server for the web browser to access the network.


pavelfeldman:
What's your scenario?

adamhj:
> What's your scenario?

for any case where user need a proxy to be configured on the tested browser, for example:
1. the application is deployed in an environment where require a proxy to access the internet
2. user want to use a man-in-the-mid proxy to monitor/modify the network traffic

pavelfeldman:
Let's file an issue for this and wait for some community feedback.

PR #284, opened 2025-04-28
chore: split context.ts into files

pavelfeldman:


src/config.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  return {
+    browserName,
+    userDataDir,
+    launchOptions,
+    contextOptions,
+  };
+}
+
+async function createUserDataDir(browserName: 'chromium' | 'firefox' | 'webkit') {
```
yury-s, line 82: Doesn't look like a part of config, move to context?

src/server.ts
```diff
@@ -51,10 +48,6 @@ export function createServerWithTools(options: Options): Server {
     };
   });
 
-  server.setRequestHandler(ListResourcesRequestSchema, async () => {
```
yury-s, line 54: Where did this and ReadResourceRequestSchema go?

yury-s (approved):

src/server.ts
pavelfeldman: We don't have resources yet.

src/config.ts
pavelfeldman: Default userDataDir resolution is a part of the config.

src/config.ts
```diff
@@ -0,0 +1,86 @@
...
+    (launchOptions as any).webSocketPort = await findFreePort();
+
+  return {
+    browserName,
+    launchOptions,
+    contextOptions,
+  };
+}
+
+async function findFreePort() {
```
yury-s, line 77: What does it have to do with config?

yury-s (approved):

Issue #272, opened 2025-04-26
Feature request: Ability to configure Playwright environment


evmin:
As a developer writing automation software I would like to be able to explicitly configure Playwright environment.

For example - to configure CORS bypass I need to set bypassCSP option and --disable-web-security launch option.

ogadra:
It would be great if it could support the Chromium command-line options listed here:
https://peter.sh/experiments/chromium-command-line-switches/

I believe a format like this would work well:

```shell
--playwright-configure option-a option-b
```

This could be easily implemented by using a [variadic option](https://www.npmjs.com/package/commander#user-content-variadic-option), allowing multiple flags to be passed through the CLI.

ogadra:
I'm imprementing now.

ogadra:
@pavelfeldman I tried it out and it works perfectly. Thank you very much!

PR #287, opened 2025-04-29
chore: allow configuring raw Playwright options

pavelfeldman:
Fixes: https://github.com/microsoft/playwright-mcp/issues/272

### DIFF
--- a/config.d.ts
+++ b/config.d.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import type * as playwright from 'playwright';
+
 export type ToolCapability = 'core' | 'tabs' | 'pdf' | 'history' | 'wait' | 'files' | 'install';
 
 export type Config = {
@@ -24,12 +26,7 @@ export type Config = {
     /**
      * The type of browser to use.
      */
-    type?: 'chrome' | 'chrome-beta' | 'chrome-canary' | 'chrome-dev' | 'chromium' | 'msedge' | 'msedge-beta' | 'msedge-canary' | 'msedge-dev' | 'firefox' | 'webkit';
-
-    /**
-     * Path to a custom browser executable.
-     */
-    executablePath?: string;
+    browserName?: 'chromium' | 'firefox' | 'webkit';
 
     /**
      * Path to a user data directory for browser profile persistence.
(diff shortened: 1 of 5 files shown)


src/tools/install.ts
```diff
@@ -30,8 +29,7 @@ const install = defineTool({
   },
 
   handle: async context => {
-    const browserOptions = await toBrowserOptions(context.config);
-    const channel = browserOptions.launchOptions?.channel ?? browserOptions.browserName ?? 'chrome';
+    const channel = context.config.browser?.launchOptions?.channel ?? context.config.browser?.launchOptions.channel ?? context.config.browser?.launchOptions.browserName ?? 'chrome';
```
yury-s, line 32:
```suggestion
    const channel = context.config.browser?.launchOptions?.channel ?? context.config.browser?.launchOptions.browserName ?? 'chrome';
```

src/config.ts
```diff
@@ -89,9 +119,57 @@ async function findFreePort() {
...
 }
 
+async function loadConfig(configFile: string | undefined): Promise<Config> {
+  if (!configFile)
+    return {};
+
+  try {
+    return JSON.parse(await fs.promises.readFile(configFile, 'utf8'));
+  } catch (error) {
+    throw new Error(`Failed to load config file: ${configFile}`);
```
yury-s, line 129: Add the error message too?

config.d.ts
```diff
@@ -24,22 +26,27 @@ export type Config = {
     /**
      * The type of browser to use.
      */
-    type?: 'chrome' | 'chrome-beta' | 'chrome-canary' | 'chrome-dev' | 'chromium' | 'msedge' | 'msedge-beta' | 'msedge-canary' | 'msedge-dev' | 'firefox' | 'webkit';
+    browserName?: 'chromium' | 'firefox' | 'webkit';
 
     /**
-     * Path to a custom browser executable.
+     * Path to a user data directory for browser profile persistence.
```
yury-s, line 32: ... If not specified, a temporary directory will be used.

src/config.ts
```diff
@@ -21,19 +21,46 @@ import path from 'path';
...
+  config?: string;
 };
 
-export async function toBrowserOptions(config: Config): Promise<BrowserOptions> {
+const defaultConfig: Config = {
+  browser: {
+    browserName: 'chromium',
+    userDataDir: os.tmpdir(),
+    launchOptions: {
+      channel: 'chrome',
```
yury-s, line 45: If the client specifies only browserName, we'll still use 'chrome' instead of 'chromium'.

yury-s (approved):

src/config.ts
pavelfeldman: That's as designed, user is in full control.

src/config.ts
pavelfeldman: Following up.

PR #290, opened 2025-04-29
docs: Fix the default path to User data directory

zwmmm:
Fix the default path to User data directory


pavelfeldman:
We should rather fix the code as we probably don't want to mix the ms-edge and chrome profiles.

yury-s (approved):

PR #297, opened 2025-04-29
chore: store channel profiles separately

pavelfeldman:


Skn0tt (approved):

src/config.ts
```diff
@@ -141,7 +141,7 @@ async function createUserDataDir(browserName: 'chromium' | 'firefox' | 'webkit')
     cacheDirectory = process.env.LOCALAPPDATA || path.join(os.homedir(), 'AppData', 'Local');
   else
     throw new Error('Unsupported platform: ' + process.platform);
-  const result = path.join(cacheDirectory, 'ms-playwright', `mcp-${browserName}-profile`);
+  const result = path.join(cacheDirectory, 'ms-playwright', `mcp-${options.channel ?? options.browserName}-profile`);
```
yury-s, line 144: This will break some of the existing clients. Do we want to do anything about it?

yury-s (approved):

src/config.ts
pavelfeldman: Not really, I think it is fine.

Issue #302, opened 2025-04-30
Migrate to ESM


mxschmitt:
We should be easily able to migrate this repository to an ESM only package. This allows us to use ESM only packages like `'open'`.

PR #303, opened 2025-04-30
chore: migrate to ESM

mxschmitt:
- [Why do I need `.js` extension?](https://stackoverflow.com/a/77150985/6512681)
- [Why setting `rootDir` in the `tsconfig.json`?](https://stackoverflow.com/a/58941798/6512681)
- [How to ensure that we add the `.js` extension via ESLint](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md#importextensions)

Fixes https://github.com/microsoft/playwright-mcp/issues/302


pavelfeldman:
> [Why do I need .js extension?](https://stackoverflow.com/a/77150985/6512681)

Let's use esbuild instead?

mxschmitt:
> > [Why do I need .js extension?](https://stackoverflow.com/a/77150985/6512681)
> 
> Let's use esbuild instead?

As per offline discussion, ESBuild doesn't support it either. Adding `.js` is common practise looking e.g. at [VSCode](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/contrib/testing/common/testCoverageService.ts#L6).

pavelfeldman (approved):

PR #308, opened 2025-04-30
test: fix 'reuse first tab when navigating' test on macOS

mxschmitt:
This fixes the following which was appearing when running `npm run ctest --workers=1  tests/tabs.spec.ts:144` on my mac machine. We have the same switch in our upstream switches.


mxschmitt:
Fixed in https://github.com/microsoft/playwright-mcp/commit/69703cc882678351e2c6e53270e14618b2199d28 already.

PR #312, opened 2025-04-30
chore: fix test harness, close the client

pavelfeldman:


tests/fixtures.ts
```diff
@@ -73,7 +73,7 @@ export const test = baseTest.extend<TestFixtures, WorkerFixtures>({
         command: 'node',
         args: [path.join(__dirname, '../cli.js'), ...args],
       });
-      const client = new Client({ name: 'test', version: '1.0.0' });
```
yury-s, line 76: Check if it doesn't exist yet

mxschmitt: Or make it an array?

pavelfeldman: We don't support multiclient mode and this PR is not trying to add one.

dgozman (approved):

PR #314, opened 2025-04-30
Add a fullPage param to the playwright_screen_capture tool

makseem77:


makseem77:

src/tools/screen.ts
```diff
@@ -28,12 +28,19 @@ const screenshot = defineTool({
...
 
-  handle: async context => {
+  handle: async (context, params) => {
     const tab = await context.ensureTab();
-    const options = { type: 'jpeg' as 'jpeg', quality: 50, scale: 'css' as 'css' };
+    const options = {
+      type: 'jpeg' as 'jpeg',
+      quality: 50,
+      scale: 'css' as 'css',
+      fullPage: params.fullPage || false,
```
Skn0tt, line 42: We use `options` to generate the code snippet below, and i'd prefer this not to list `fullPage: undefined` in the common case. Could we try only adding this if `params.fullPage` is true?

Skn0tt (commented):
The change looks good, left a small comment. Out of curiosity, could you share your usecase for this? Why does it matter to you? Also, have you experimented with what happens if you always set `fullPage: true`?

src/tools/screen.ts
makseem77: :+1:

README.md
```diff
@@ -326,7 +326,8 @@ X Y coordinate space, based on the provided screenshot.
 
 - **browser_screen_capture**
   - Description: Take a screenshot of the current page
-  - Parameters: None
+  - Parameters:
+    - `fullPage` (boolean, optional): Whether the whole page should be screenshotted, or only the current viewport
```
pavelfeldman, line 330: `fullPage` is never a good idea for automation or a page outside of your control. Most of the time the screenshot will be incomplete due to memory issues.

makseem77: I did not have any issues testing, but it was only on controlled pages of reasonable size.

demostanis: the page is in our control

pavelfeldman: > I did not have any issues testing, but it was only on controlled pages of reasonable size.

I think adding it would be misleading as it will rarely work in practice. Even if we manage to capture a meaningful fullPage screenshot, llm won't handle the size.

makseem77: It seems to work perfectly on my side for regular sized web pages, and the LLM can read and describe the full image without any issue.
Why should we have it only in playwright but not in the playwright-mcp ?

ausangshukla:
I need this too, but somehow the AI does not take a fullpage screenshot, is this available in latest?

loia5tqd001:
I also don't think there's a good reason not to support this, please reopen it. If it doesn't work for a user with a too long page, it's up to them, while it might work for a lot of other cases @pavelfeldman

ausangshukla:
Please can someone confirm if this can be added, really need the full page screenshot, and not sure how to fork and add my own, so would be great to see it in the main branch

makseem77:
@ausangshukla i have a fork supporting it on my repository

ausangshukla:
> @ausangshukla i have a fork supporting it on my repository

I was not able to run your fork, would you be able to provide a one liner on how to run the fork. TIA

makseem77:
> I also don't think there's a good reason not to support this, please reopen it. If it doesn't work for a user with a too long page, it's up to them, while it might work for a lot of other cases @pavelfeldman

@Skn0tt can this be reopened ?

Skn0tt:
Please open an issue so we can discuss the need for this, before we discuss implementations.

PR #324, opened 2025-05-02
fix: sticky launch errors

mxschmitt:
This fixes an issue that there were sticky launch errors. When the [following code path](https://github.com/microsoft/playwright-mcp/blob/a15f0f301b405c00e9987fd84bf0f1f1fe856125/src/context.ts#L307-L339) was throwing, the Error was stored in the Promise and not cleared afterwards, this meant:

- If a browser was not there and the user tried to install it via `browser_install` it was never working since the error was sticky.
- If other errors like CDP is not available yet etc. error appear a re-connect would not work - the MCP server would require a restart.

Test plan: Since we don't have any `browser_install` tests I added a CDP test for now to cover this bug.

### DIFF
--- a/tests/cdp.spec.ts
+++ b/tests/cdp.spec.ts
@@ -17,7 +17,7 @@
 import { test, expect } from './fixtures.js';
 
 test('cdp server', async ({ cdpEndpoint, startClient }) => {
-  const client = await startClient({ args: [`--cdp-endpoint=${cdpEndpoint}`] });
+  const client = await startClient({ args: [`--cdp-endpoint=${await cdpEndpoint()}`] });
   expect(await client.callTool({
     name: 'browser_navigate',
     arguments: {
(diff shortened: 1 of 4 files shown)


tests/cdp.spec.ts
```diff
@@ -54,3 +54,21 @@ test('cdp server reuse tab', async ({ cdpEndpoint, startClient }) => {
 \`\`\`
 `);
 });
+
+test('should throw connection error and allow re-connecting', async ({ cdpEndpoint, startClient }) => {
+  const port = 3200 + (+process.env.TEST_PARALLEL_INDEX!);
```
Skn0tt, line 59: I know it's existed before, but what the heck is this env var? Shouldn't it be `testInfo.parallelIndex`?

Skn0tt (approved):
Looks good. Let's figure out what the heck is that env var though.

tests/cdp.spec.ts
mxschmitt: Yes.

Skn0tt (approved):

PR #327, opened 2025-05-02
chore: don't sanitize file extension away

Skn0tt:


src/tools/utils.ts
```diff
@@ -70,6 +70,13 @@ export async function waitForCompletion<R>(context: Context, page: playwright.Pa
   }
 }
 
-export function sanitizeForFilePath(s: string) {
+function sanitize(s: string) {
```
pavelfeldman, line 73: It should keep the name as `sanitize` in utils is meaningless.

pavelfeldman (approved):
Lgtm mod comments

PR #328, opened 2025-05-02
fix: don't error on navigating to a download link

Skn0tt:


src/tab.ts
```diff
@@ -66,7 +66,28 @@ export class Tab {
   }
 
   async navigate(url: string) {
-    await this.page.goto(url, { waitUntil: 'domcontentloaded' });
+    const downloadEvent = this.page.waitForEvent('download');
+    try {
+      await this.page.goto(url, { waitUntil: 'domcontentloaded' });
+    } catch (e: unknown) {
+      if (!(e instanceof Error))
```
pavelfeldman, line 73: We never throw non-errors in Playwright API.

src/tab.ts
```diff
@@ -66,7 +66,28 @@ export class Tab {
...
+        throw e;
+
+      const mightBeDownload =
```
pavelfeldman, line 76: Do we know if the download event has been fired? It would be easier to understand if you were listening to the download event, and if it was emitted (make sure to listen on the context not on a page), you would wait for the download.

Skn0tt: That's what the code is doing, it's not waiting for it yet though. The download event kicks off only after the error though, so we have to check the error messages so we don't add 500ms of wait to every navigate call.

Skn0tt: as discussed yesterday, let's move forward with this. i'll open an upstream issue about making this discoverable from chromium.

pavelfeldman (approved):

PR #349, opened 2025-05-05
chore: support custom filename in screenshot function

ogadra:
## Description

This PR supports specifying custom filenames for screenshots programmatically.
This allows for more descriptive filenames, making it easier to identify the context or content of a screenshot without needing to open the file, particularly when multiple screenshots are captured in sequence.

## Motivation

The current timestamp-based filenames make it hard to identify specific screenshots when multiple are generated. Providing meaningful, custom filenames (e.g., `login_page_error.png` or `user_dashboard_loaded.png`) significantly improves organization and debugging efficiency, as the purpose of each image is clear from its name.

## Changes

- Added an optional `filename` argument to the `browser_take_screenshot` function.


tests/screenshot.spec.ts
```diff
@@ -107,16 +107,112 @@ test('browser_take_screenshot (outputDir)', async ({ startClient }, testInfo) =>
...
+  })).toEqual({
+    content: [
+      {
+        data: expect.any(String),
+        mimeType: 'image/jpeg',
+        type: 'image',
+      },
+      {
+        text: expect.stringMatching(
+            new RegExp(`// Screenshot viewport and save it as ${testInfo.outputPath('output')}/page-\\d{4}-\\d{2}-\\d{2}T\\d{2}-\\d{2}-\\d{2}\\-\\d{3}Z\\.\\jpeg`)
```
ogadra, line 122: Verify timestamp filename presence.

src/tools/snapshot.ts
```diff
@@ -208,6 +208,7 @@ const selectOption = defineTool({
 
 const screenshotSchema = z.object({
   raw: z.boolean().optional().describe('Whether to return without compression (in PNG format). Default is false, which returns a JPEG image.'),
+  filename: z.string().optional().describe('File name to save the screenshot to. Defaults to `page-${new Date().toISOString()}.${fileType}` if not specified.'),
```
pavelfeldman, line 211: Keep it simpler for LLM to fllow
```suggestion
  filename: z.string().optional().describe('File name to save the screenshot to. Defaults to `page-{timestamp}.{png|jpeg}` if not specified.'),
```

tests/screenshot.spec.ts
pavelfeldman: Windows might be unhappy about your /

tests/screenshot.spec.ts
```diff
@@ -107,16 +107,112 @@ test('browser_take_screenshot (outputDir)', async ({ startClient }, testInfo) =>
...
+  const files = [...fs.readdirSync(outputDir)];
+
   expect(fs.existsSync(outputDir)).toBeTruthy();
-  expect([...fs.readdirSync(outputDir)]).toHaveLength(1);
+  expect(files).toHaveLength(1);
+  expect(files[0]).toMatch(/^page-\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}-\d{3}Z\.jpeg$/);
 });
 
-test('browser_take_screenshot (omitBase64)', async ({ startClient }) => {
+test('browser_take_screenshot (raw: true)', async ({ startClient }, testInfo) => {
```
pavelfeldman, line 136: These are very similar, you can generate them via for (const raw of [undefined, true])

pavelfeldman (commented):
While you are here, do that for the pdf as well.

tests/screenshot.spec.ts
ogadra: We assume that the full path is already verified in subsequent tests. In this test, we verify that the response contains the file name.

src/tools/snapshot.ts
ogadra: I'll also update README.md to reflect this change.

pavelfeldman (approved):

pavelfeldman:
i'll follow up with the fix - our docker tests are picky.

PR #353, opened 2025-05-06
Introduce tool for generating reusable element locators

ruichu:
**Problem**:
Currently, element references obtained via browser_take_screenshot are transient. They become invalid after a page refresh or across different browser sessions, requiring the Agent to re-capture a screenshot to interact with the same logical element again.

**Solution**:
This PR introduces a new tool that generates a reusable locator for a given element.

**Benefits**:
Provides the Agent with a persistent way to identify elements across page loads and sessions.
Eliminates the need for repeated calls to browser_take_screenshot solely for element identification.
Improves the robustness and efficiency of agent interactions over time.

### DIFF
--- a/src/tools/snapshot.ts
+++ b/src/tools/snapshot.ts
@@ -289,7 +289,7 @@ const getLocator = defineTool({
           content: [{ type: 'text', text: await generateLocator(locator) }]
         };
       },
-      captureSnapshot: true,
+      captureSnapshot: false,
       waitForNetwork: false,
     };
   },
--- a/src/tools/snapshot.ts
+++ b/src/tools/snapshot.ts
@@ -265,6 +265,36 @@ const screenshot = defineTool({
   }
 });
 
+const getLocator = defineTool({
+  capability: 'core',
+  schema: {
+    name: 'browser_get_locator',
+    description: 'Get reusable locator for the element',
+    inputSchema: elementSchema,
+  },
+
+  handle: async (context, params) => {
+    const tab = context.currentTabOrDie();
+    const locator = tab.snapshotOrDie().refLocator(params.ref);
+
+    const code = [
+      `// Get reusable locator for ${params.element}`,
+      `await page.${await generateLocator(locator)};`
+    ];
+
+    return {
+      code,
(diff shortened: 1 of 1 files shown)


Skn0tt:
Hi @ruichu! What's your scenario for this?

ruichu:
> Hi @ruichu! What's your scenario for this?

I want the Agent to generate reusable UI automation code, rather than just running UI automation operations once. In this case, I cannot use refs from snapshots as locators, but must use locators that are only related to the UI structure.

Skn0tt:
In my experience, the LLM was able to infer a proper locator based on the ARIA snapshot very well. Could you share a prompt that you're using where it doesn't work for you?

ruichu:
> In my experience, the LLM was able to infer a proper locator based on the ARIA snapshot very well. Could you share a prompt that you're using where it doesn't work for you?

For example, 

```
You are a UI automation programmer.
You are given a task to write Python code snippet to automate a web page using Playwright.
The code should only include Playwright API calls, and should not include any other code.
You can use the tools provided to help you write the code. 
For example, you can use the tool to inspect the structure of the web page, or get the selectors of the web elements.
You can even use the tool to interact with the web page if necessary.
```

Such a prompt doesn't work because the Playwright code it generates uses locators like "ref=s1e37". These locators are only valid for the current snapshot, and if you try to run the generated Playwright code, it won't work.

When I add this tool, I will tell the LLM in the prompt:

```
Notice: The locator of the web page snapshot is NOT stable, so you should get the reusable locator of the web element using the tool "browser_get_locator", and use the reusable locator in your code.
```

This way, it will generate code similar to the following. As long as the web structure doesn't change, the following code will always work, and it no longer depends on the LLM when running.

```python
    page.locator('#first-name').fill("John")
    page.locator('#last-name').fill("Doe")
```

ruichu:
The term 'selector' in my prompt should be 'locator', to adapt to Playwright's naming convention.

pavelfeldman:
I'm not sure I follow the rationale or how the tool is used. Please start with filing an issue and showing example MCP client trajectory that requires the new tool.

Issue #346, opened 2025-05-04
browser_navigate does not work in docker image


jsilhan:
When trying via MCP inspect "docker run -i --rm --init mcp/playwright" and browser_navigate tool I get an error:
> Error: Browser specified in your config is not installed. Either install it (likely) or change the config.

![Image](https://github.com/user-attachments/assets/c1bd206c-4959-4f90-a863-e1387298d5ff)

Docker should be already configured for headless chromium according to the README.
I have tried to pull the `mcp/playwright` image from docker registry and also build it from the repo but neither solution worked.

I can't use npx version as I am running on Fedora Linux.

cmrigney:
There's been a regression. See the comments here: https://github.com/microsoft/playwright-mcp/pull/273

pavelfeldman:
I landed the change. I think we should start pushing docker images with the releases inside our pipeline. That'll be the easiest.

PR #361, opened 2025-05-06
test: run tests on MCP server inside Docker

mxschmitt:
https://github.com/microsoft/playwright-mcp/issues/346

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -29,6 +29,7 @@ export default defineConfig<TestOptions>({
     { name: 'chrome' },
     { name: 'msedge', use: { mcpBrowser: 'msedge' } },
     { name: 'chromium', use: { mcpBrowser: 'chromium' } },
+    ...process.env.MCP_IN_DOCKER ? [{ name: 'chromium-docker', use: { mcpBrowser: 'chromium', mcpMode: 'docker' as const } }] : [],
     { name: 'firefox', use: { mcpBrowser: 'firefox' } },
     { name: 'webkit', use: { mcpBrowser: 'webkit' } },
   ],
--- a/tests/config.spec.ts
+++ b/tests/config.spec.ts
@@ -19,7 +19,8 @@ import fs from 'node:fs';
 import { Config } from '../config.js';
 import { test, expect } from './fixtures.js';
 
-test('config user data dir', async ({ startClient, mcpBrowser }, testInfo) => {
+test('config user data dir', async ({ startClient, mcpMode }, testInfo) => {
+  test.skip(mcpMode === 'docker', 'Mounting files is not supported in docker mode');
   const config: Config = {
     browser: {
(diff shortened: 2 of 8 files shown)


playwright.config.ts
```diff
@@ -29,6 +29,7 @@ export default defineConfig<TestOptions>({
     { name: 'chrome' },
     { name: 'msedge', use: { mcpBrowser: 'msedge' } },
     { name: 'chromium', use: { mcpBrowser: 'chromium' } },
+    ...process.platform === 'linux' ? [{ name: 'chromium-docker', use: { mcpBrowser: 'chromium', mcpMode: 'docker' as const } }] : [],
```
pavelfeldman, line 32: Nit: we should follow up and split this into multiple configs.

tests/fixtures.ts
```diff
@@ -124,12 +121,22 @@ export const test = baseTest.extend<TestFixtures & TestOptions, WorkerFixtures>(
     browserProcess?.kill();
   },
 
+  rewriteFile: async ({ mcpMode }, use) => {
+    await use((file: string) => {
+      if (mcpMode === 'docker')
+        return path.relative(path.dirname(test.info().config.configFile!), file);
```
pavelfeldman, line 127: I think this surfaces a real issue - users that run Playwright MCP server in Docker don't expect the files to be generated in Docker. I wonder if we should disable all the file saving in that mode and expose those files as resources in the future.

What I'm suggesting: do not run tests that require this rebase since this is not what user expects.

tests/fixtures.ts
```diff
@@ -156,6 +156,22 @@ export const test = baseTest.extend<TestFixtures & TestOptions, WorkerFixtures>(
...
+  // NOTE: Can be removed when we drop Node.js 18 support and changed to import.meta.filename.
+  const __filename = url.fileURLToPath(import.meta.url);
+  if (mcpMode === 'docker') {
+    const dockerArgs = ['run', '--rm', '-i', '--network=host', '-v', `${test.info().project.outputDir}:/app/test-results`];
+    return new StdioClientTransport({
+      command: 'docker',
+      args: [...dockerArgs, 'playwright-mcp-dev:latest', ...args],
+    });
+  }
+  return new StdioClientTransport({
```
pavelfeldman, line 169: We now must provide cwd since we read files relative to it.

tests/config.spec.ts
```diff
@@ -19,7 +19,8 @@ import fs from 'node:fs';
 import { Config } from '../config.js';
 import { test, expect } from './fixtures.js';
 
-test('config user data dir', async ({ startClient, mcpBrowser }, testInfo) => {
+test('config user data dir', async ({ startClient, mcpMode }, testInfo) => {
```
pavelfeldman, line 22: Introducing a `localOutputPath()` fixture that skips automatically will make all these tests nicer.

PR #363, opened 2025-05-07
Install Playwright browser dependencies

JOBKO-QA:
Please disregard that — it was uploaded by mistake.


mxschmitt:
Why is this needed? Our tests are passing without that. Could you elaborate and/or file an issue?

PR #380, opened 2025-05-09
docs: reference to new Docker image

mxschmitt:
See [Microsoft MCR](https://mcr.microsoft.com/en-us/artifact/mar/playwright/mcp/about) for more information - not sure if its helpful to link to this page from the README?

Closes https://github.com/microsoft/playwright-mcp/issues/346

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ And then in MCP client config, set the `url` to the SSE endpoint:
   "mcpServers": {
     "playwright": {
       "command": "docker",
-      "args": ["run", "-i", "--rm", "--init", "mcr.microsoft.com/playwright/mcp"]
+      "args": ["run", "-i", "--rm", "--init", "--pull=always", "mcr.microsoft.com/playwright/mcp"]
     }
   }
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,7 +207,7 @@ And then in MCP client config, set the `url` to the SSE endpoint:
   "mcpServers": {
     "playwright": {
       "command": "docker",
-      "args": ["run", "-i", "--rm", "--init", "mcp/playwright"]
+      "args": ["run", "-i", "--rm", "--init", "mcr.microsoft.com/playwright/mcp"]
```
mxschmitt, line 210: If we want to ensure that users always run the latest version of the image, we can add `"--pull=always"` — similar to using `@latest` in npm. Docker will check if a newer version is available and only pull it if needed.
 
```suggestion
      "args": ["run", "-i", "--rm", "--init", "--pull=always", "mcr.microsoft.com/playwright/mcp"]
```

pavelfeldman (approved):

Issue #382, opened 2025-05-09
createConnection rewrite break the programe


ssddi456:
https://github.com/microsoft/playwright-mcp/blob/57b3c14276a00f9e8887eb304c9ffc90b2aa5c7a/src/index.ts#L22

this change made a selfref call which break the function

should be a reexport of ?

https://github.com/microsoft/playwright-mcp/blob/57b3c14276a00f9e8887eb304c9ffc90b2aa5c7a/src/connection.ts#L27

PR #384, opened 2025-05-09
fix: createConnection() via public API

mxschmitt:
Fixes https://github.com/microsoft/playwright-mcp/issues/382

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ http.createServer(async (req, res) => {
   // ...
 
   // Creates a headless Playwright MCP server with SSE transport
-  const connection = await createConnection({ headless: true });
+  const connection = await createConnection({ browser: { launchOptions: { headless: true } } });
   const transport = new SSEServerTransport('/messages', res);
   await connection.connect(transport);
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -231,7 +231,7 @@ http.createServer(async (req, res) => {
   // ...
 
   // Creates a headless Playwright MCP server with SSE transport
-  const connection = await createConnection({ headless: true });
+  const connection = await createConnection({ browser: { launchOptions: { headless: true } } });
```
mxschmitt, line 234: This is supposed to be a SSE transport so I start wondering if we should adopt something similar here in this example like we have in the tests: `sessions` map and GET/POST handling, otherwise it didn't work for me.

src/index.ts
```diff
@@ -14,10 +14,10 @@
...
  */
 
-import { Connection } from './connection.js';
+import { Connection, createConnectionImpl  } from './connection.js';
 
 import type { Config } from '../config.js';
 
 export async function createConnection(config: Config = {}): Promise<Connection> {
-  return createConnection(config);
+  return createConnectionImpl(config);
```
pavelfeldman, line 22: I think the bug here is that I forgot to import createConnection from connection, no other changes should be necessary.

pavelfeldman (approved):

PR #385, opened 2025-05-09
fix: use of wrong launchOptions type in public API

mxschmitt:


config.d.ts
```diff
@@ -40,7 +40,7 @@ export type Config = {
      *
      * This is useful for settings options like `channel`, `headless`, `executablePath`, etc.
      */
-    launchOptions?: playwright.BrowserLaunchOptions;
+    launchOptions?: playwright.LaunchOptions;
```
dgozman, line 43: How come TSC did not complain? Are we missing tsc linter on the bots?

dgozman (approved):

config.d.ts
mxschmitt: `/config.d.ts` is not included in the `tsconfig.json` - only `src` is.

mxschmitt: https://github.com/microsoft/playwright-mcp/pull/386

Issue #389, opened 2025-05-09
Can you add implementation to wait for a object to get visible or disappear


mayur-waghela:
Lets say i want to wait until a button is visible with max wait time being 10 seconds
Or the other way wait until the loading message is gone.

For now i am explicitly waiting for 5 or 10 seconds for the object to appear.. this feature will increase the speed of execution

PR #390, opened 2025-05-09
feat(wait): allow waiting for given text

pavelfeldman:
Fixes https://github.com/microsoft/playwright-mcp/issues/389

### DIFF
--- a/src/tools/common.ts
+++ b/src/tools/common.ts
@@ -21,19 +21,44 @@ const wait: ToolFactory = captureSnapshot => defineTool({
   capability: 'wait',
 
   schema: {
-    name: 'browser_wait',
-    title: 'Wait',
-    description: 'Wait for a specified time in seconds',
+    name: 'browser_wait_for',
+    title: 'Wait for',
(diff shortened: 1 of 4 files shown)


src/tools/common.ts
```diff
@@ -21,19 +21,44 @@ const wait: ToolFactory = captureSnapshot => defineTool({
...
 
   handle: async (context, params) => {
-    await new Promise(f => setTimeout(f, Math.min(10000, params.time * 1000)));
+    if (!params.text && !params.textGone && !params.time)
+      throw new Error('Either time, text or textGone must be provided');
+
+    const code: string[] = [];
+
+    if (params.time) {
+      code.push(`await new Promise(f => setTimeout(f, ${params.time!} * 1000));`);
```
yury-s, line 42: Print actual computed timeout?

yury-s (approved):

src/tools/common.ts
pavelfeldman: I like it more this way!

PR #396, opened 2025-05-11
Fix import in README from createServer to createConnection

ta1m1kam:
Probably, `createServer` is not from `@playwright/mcp`.


Skn0tt:
cc @mxschmitt - looks like this changed?

pavelfeldman (approved):

PR #413, opened 2025-05-13
chore: normalize --no- options

pavelfeldman:


tests/fixtures.ts
```diff
@@ -65,6 +65,8 @@ export const test = baseTest.extend<TestFixtures & TestOptions, WorkerFixtures>(
 
     await use(async options => {
       const args = ['--user-data-dir', path.relative(configDir, userDataDir)];
+      if (process.env.CI && process.platform === 'linux')
```
yury-s, line 68: Don't our users need the same?

src/config.ts
```diff
@@ -111,8 +111,10 @@ export async function configFromCLIOptions(cliOptions: CLIOptions): Promise<Conf
 
   if (browserName === 'chromium') {
     (launchOptions as any).cdpPort = await findFreePort();
-    if (cliOptions.sandbox === false)
+    if (!cliOptions.sandbox) {
```
yury-s, line 114: isn't it undefined by default?

yury-s (approved):

src/config.ts
pavelfeldman: it is not.

tests/fixtures.ts
pavelfeldman: Only those that run MCP on CI (nobody)

PR #423, opened 2025-05-14
fix: Expand `browser_type` description with exact elements it can interact with

mastrzyz:
Give AI a hint of what elements it can type into to reduce errors such as below:

```
 [{"type":"text","text":"Error: locator.fill: Error: Element is not an <input>, <textarea> or [contenteditable] element\nCall log:\n\u001b[2m  - waiting for locator('aria-ref=s23e585')\u001b[22m\n\u001b[2m    - locator resolved to <div role=\"combobox\" spellcheck=\"false\" aria-expanded=\"true\" aria-haspopup=\"listbox\" aria-owns=\"downshift-0-menu\" aria-labelledby=\"people-picker-label\" class=\"ui-box ac fl fk l fu en nu nv am nw nx ny ui-input ac fl fk l fu en nu ui-dropdown__searchinput__wrapper ui-dropdown__searchinput\">…</div>\u001b[22m\n\u001b[2m    - fill(\"testuser@example.com\")\u001b[22m\n\u001b[2m  - attempting fill action\u001b[22m\n\u001b[2m    - waiting for element to be visible, enabled and editable\u001b[22m\n"}]
```

### DIFF
--- a/src/tools/snapshot.ts
+++ b/src/tools/snapshot.ts
@@ -147,7 +147,7 @@ const type = defineTool({
   schema: {
     name: 'browser_type',
     title: 'Type text',
-    description: 'Type text into editable element',
+    description: 'Type text into editable element such as <input>, <textarea> and [contenteditable] elements',
     inputSchema: typeSchema,
     type: 'destructive',
   },


pavelfeldman:
That's a good error, is it not recovering from it?

mastrzyz:
> That's a good error, is it not recovering from it?

yes if we sent it back to AI but this is 2 AI Requests vs potentially one if AI was better informed?

pavelfeldman:
Using more tokens for the tool description degrades overall performance for all tasks. So if this is a recoverable edge case, returning error is more efficient.

mastrzyz:
> Using more tokens for the tool description degrades overall performance for all tasks. So if this is a recoverable edge case, returning error is more efficient.

true, we need data on how often this occurs and how much retrying costs in tokens vs what the token cost of adding this is.


It probably also depends on the subject under test, our clients AOM is huge so probably gets confused more frequently..

pavelfeldman:
Closing for now

PR #427, opened 2025-05-15
chore: start trace server

pavelfeldman:


src/transport.ts
```diff
@@ -140,6 +140,6 @@ export function startHttpTransport(config: FullConfig, port: number, hostname: s
       'If your client supports streamable HTTP, you can use the /mcp endpoint instead.',
     ].join('\n');
     // eslint-disable-next-line no-console
-    console.log(message);
+    console.error(message);
```
yury-s, line 143: This will break existing users.

yury-s (approved):

src/transport.ts
pavelfeldman: Unlikely, servers are not started programmatically.

Issue #439, opened 2025-05-16
Playwright-MCP 0.0.22 is causing " Execution context was destroyed, most likely because of a navigation" failures with an existing browser


mastrzyz:
Updating @playwright/mcp from `0.0.20` to `0.0.22`

Gives us the following error
```
Error: MCP Tool browser_snapshot failed with error: [{"type":"text","text":"Error: page._snapshotForAI: Execution context was destroyed, most likely because of a navigation"}]
```

The Tool with this error is `browser_snapshot`

We start the @playwright/mcp programatically with an already running browser : 

```ts
  const server = await createConnection({
    browser: {
      cdpEndpoint: `http://localhost:${CDP_PORT}`,
    },
  });
```

mastrzyz:
any clue what changed between 20 and 22? anything I can try to get more logs here.

pavelfeldman:
If there is a scenario we can run locally, happy to take a look.

mastrzyz:
> If there is a scenario we can run locally, happy to take a look.


Here is the workflow : 

1. Playwright test runs and spawns the browser, logs in to teams etc. 

Here is how Playwright starts the browser

```
[...]

PR #440, opened 2025-05-17
chore: fix operation over cdp

pavelfeldman:
Ref https://github.com/microsoft/playwright-mcp/issues/439

### DIFF
--- a/src/pageSnapshot.ts
+++ b/src/pageSnapshot.ts
@@ -40,6 +40,9 @@ export class PageSnapshot {
   }
 
   private async _build() {
+    // FIXME: Rountrip evaluate to ensure _snapshotForAI works.
+    // This probably broke once we moved off locator snapshots
+    await this._page.evaluate(() => 1);
     const snapshot = await callOnPageNoTrace(this._page, page => (page as PageEx)._snapshotForAI());
     this._text = [
       `- Page Snapshot`,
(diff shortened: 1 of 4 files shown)


tests/fixtures.ts
```diff
@@ -95,39 +100,25 @@ export const test = baseTest.extend<TestFixtures & TestOptions, WorkerFixtures>(
...
-    await use(async port => {
-      if (!port)
-        port = 3200 + test.info().parallelIndex;
-      if (browserProcess)
-        return `http://localhost:${port}`;
-      browserProcess = spawn(chromium.executablePath(), [
-        `--user-data-dir=${testInfo.outputPath('user-data-dir')}`,
-        `--remote-debugging-port=${port}`,
-        `--no-first-run`,
-        `--no-sandbox`,
```
yury-s, line 110: What about these options, don't we need them?

src/pageSnapshot.ts
```diff
@@ -40,6 +40,9 @@ export class PageSnapshot {
   }
 
   private async _build() {
+    // FIXME: Rountrip evaluate to ensure _snapshotForAI works.
```
yury-s, line 43: This is not reassuring 😄

yury-s (approved):

tests/fixtures.ts
pavelfeldman: launchPersistentProfile has them, and more

src/pageSnapshot.ts
pavelfeldman: you don't say

tests/fixtures.ts
mxschmitt: I think `launchPersistentContext` adds them all internally.

PR #448, opened 2025-05-19
Add 'browser_evaluate' tool for javascript execution in browser

gavyn-ezell:


gavyn-ezell:

pavelfeldman:
Not sure it makes sense for our tool!

PR #450, opened 2025-05-20
feat(cors): Add CORS policy check and response to browsers.

AOrazaev:


config.d.ts
```diff
@@ -114,6 +114,11 @@ export type Config = {
      * List of origins to block the browser to request. Origins matching both `allowedOrigins` and `blockedOrigins` will be blocked.
      */
     blockedOrigins?: string[];
+
+    /**
+     * Origins to allow by CORS policy. Can be regex.
+     */
+    corsAllowOrigins?: RegExp[];
```
pavelfeldman, line 121: This rather belongs to the 'server' section.

src/transport.ts
```diff
@@ -33,7 +33,31 @@ export async function startStdioTransport(config: FullConfig, connectionList: Co
...
 
+function checkCors(config: FullConfig, req: http.IncomingMessage, res: http.ServerResponse): boolean {
+  if (config.network?.corsAllowOrigins === undefined)
+    return false;
+
+  const origin = req.headers.origin;
+  if (origin === undefined)
+    return false;
+
+  if (config.network.corsAllowOrigins.some(re => re.test(origin))) {
```
pavelfeldman, line 44: What is your use case? This exposes universal access to the host machine to the client, including full access to the fs.

src/program.ts
```diff
@@ -34,6 +34,7 @@ program
     .option('--caps <caps>', 'comma-separated list of capabilities to enable, possible values: tabs, pdf, history, wait, files, install. Default is all.')
     .option('--cdp-endpoint <endpoint>', 'CDP endpoint to connect to.')
     .option('--config <path>', 'path to the configuration file.')
+    .option('--cors-allow-origins <origin>', 'semicolon-separated list of allowed origins by CORS policy. Can be regex.', semicolonSeparatedList)
```
pavelfeldman, line 37: It'd be nice if origins syntax was consistent (blocked, allowed, cors). Let's not support regex here.

src/transport.ts
```diff
@@ -31,7 +32,31 @@ export async function startStdioTransport(server: Server) {
...
+    res.setHeader('Access-Control-Max-Age', 2592000);
+    res.setHeader('Access-Control-Allow-Headers', 'mcp-session-id, content-type');
+    return true;
+  }
+
+  return false;
+}
+
 async function handleSSE(server: Server, req: http.IncomingMessage, res: http.ServerResponse, url: URL, sessions: Map<string, SSEServerTransport>) {
+  if (checkCors(server.config, req, res) && req.method === 'OPTIONS') {
```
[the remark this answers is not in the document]
AOrazaev: I don't think it is a good suggestion. most of functionality is already in a shared function

src/config.ts
```diff
@@ -181,6 +182,7 @@ export async function configFromCLIOptions(cliOptions: CLIOptions): Promise<Conf
     server: {
       port: cliOptions.port,
       host: cliOptions.host,
+      corsAllowedOrigins: cliOptions.corsAllowedOrigins ? cliOptions.corsAllowedOrigins : undefined,
```
pavelfeldman:
```suggestion
      corsAllowedOrigins: cliOptions.corsAllowedOrigins,
```

src/transport.ts
pavelfeldman, line 55:
```suggestion
  if (req.method === 'OPTIONS') {
    if (checkCors(server.config, req, res)) {
      ...
    } else {
      ...
    }
    return;
  }
```

src/transport.ts
```diff
@@ -76,6 +101,11 @@ async function handleStreamable(server: Server, req: http.IncomingMessage, res:
     return await transport.handleRequest(req, res);
   }
 
+  if (checkCors(server.config, req, res) && req.method === 'OPTIONS') {
```
pavelfeldman, line 104: ditto

src/transport.ts
AOrazaev: Hey Pavel, thanks for your suggestion!

Unfortunately, as far as I understand, this particular suggestion will break the behavior. Browser will require CORS headers to be populated not only for "preflight" request (OPTIONS), but also for every consequent request (POST/GET). So OPTIONS will succeed, actual POST will break because of cors.

This is the reason I've put the function check in front. I understand that this is probably not an obvious behavior and it happens implicitly. Feel free to suggest how to make it better and more explicit - I'll happily make a change.

Thank you!

src/transport.ts
AOrazaev: Answered in the main suggestion comment.

src/transport.ts
pavelfeldman:
```js
const { corsAllowed } = applyCorsHeaders(...);
if (req.method === 'OPTIONS' && corsAllowed)
  return /*204*/;
```

pavelfeldman (commented):
We usually write tests for the changes. You can consider reusing the sse tests and accessing it from the test server (server fixture).

Issue #458, opened 2025-05-22
CLI browserName is set to 'chromium' even when not specified, causing unexpected override in merged config


crabdesing:
Hi,

I have noticed a potential issue in the config resolution logic regarding the browserName setting when using CLI options.

In [`src/config.ts`](https://github.com/microsoft/playwright-mcp/blob/aa6ac51f92f199a4c50eb7c5f191452e308dfd5c/src/config.ts#L123), in the `configFromCLIOptions` function, if the CLI option for `browser` is not specified, the switch statement defaults to setting `browserName` to `'chromium'` and `channel` to `'chrome'`. As a result, when merging configs in `resolveCLIConfig`, the `cliOverrides` object includes `browserName: 'chromium'`, even though the user did not specify it via CLI.

This leads to an unexpected override:  
- If the user does not specify the browser via CLI, but a different `browserName` is set in the config file, the CLI override will still force it to `'chromium'`.

**Reference Code:**  
[`src/config.ts` - line 123](https://github.com/microsoft/playwright-mcp/blob/aa6ac51f92f199a4c50eb7c5f191452e308dfd5c/src/config.ts#L123)

**Expected behavior:**  
If no browser is specified via CLI, the merged config should respect the value from the config file, and not forcibly override it with `'chromium'`.

**Possible solution:**  
Consider updating `configFromCLIOptions` so that `browserName` is only set if the CLI option is actually provided. Otherwise, leave it undefined so that the config file value is preserved.

Thank you!

Skn0tt:
Thanks for the bug report! Can reproduce it with this test:

```ts
test('browserName', { annotation: { type: 'issue', description: 'https://github.com/microsoft/playwright-mcp/issues/458' } }, async ({ startClient, localOutputPath }) => {
  const config: Config = {
    browser: {
      browserName: 'firefox',
    },
  };
  const configPath = localOutputPath('config.json');
  await fs.promises.writeFile(configPath, JSON.stringify(config, null, 2));
  const client = await startClient({ args: ['--config', configPath] });
  expect(await client.callTool({
    name: 'browser_navigate',
    arguments: { url: 'data:text/html,<script>document.title = navigator.userAgent</script>' },
  })).toContainTextContent(`Firefox`);
});
```

PR #461, opened 2025-05-22
fix: respect browserName in config

Skn0tt:
Resolves https://github.com/microsoft/playwright-mcp/issues/458

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -128,7 +128,7 @@ export async function configFromCLIOptions(cliOptions: CLIOptions): Promise<Conf
     channel,
     executablePath: cliOptions.executablePath,
     headless: cliOptions.headless,
-    chromiumSandbox: !!cliOptions.sandbox,
+    chromiumSandbox: cliOptions.sandbox ? undefined : false, // --no-sandbox was passed, disable the sandbox
   };
 
   if (cliOptions.proxyServer) {
--- a/src/config.ts
+++ b/src/config.ts
@@ -128,8 +128,11 @@ export async function configFromCLIOptions(cliOptions: CLIOptions): Promise<Conf
     channel,
     executablePath: cliOptions.executablePath,
     headless: cliOptions.headless,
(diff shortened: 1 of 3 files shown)


tests/fixtures.ts
```diff
@@ -79,7 +79,7 @@ export const test = baseTest.extend<TestFixtures & TestOptions, WorkerFixtures>(
       if (options?.args)
         args.push(...options.args);
       if (options?.config) {
-        const configFile = testInfo.outputPath('config.json');
+        const configFile = localOutputPath('config.json');
```
pavelfeldman, line 82: You just disabled all the tests on the docker.

tests/config.spec.ts
```diff
@@ -16,29 +16,37 @@
...
-  const config: Config = {
-    browser: {
-      userDataDir: localOutputPath('user-data-dir'),
-    },
-  };
-  const configPath = localOutputPath('config.json');
-  await fs.promises.writeFile(configPath, JSON.stringify(config, null, 2));
-
-  const client = await startClient({ args: ['--config', configPath] });
+  const userDataDir = localOutputPath('user-data-dir');
```
pavelfeldman, line 26: What was wrong with this test?

src/config.ts
```diff
@@ -119,26 +121,16 @@ export async function configFromCLIOptions(cliOptions: CLIOptions): Promise<Conf
...
-      browserName = 'chromium';
-      channel = 'chrome';
   }
 
   // Launch options
   const launchOptions: LaunchOptions = {
     channel,
     executablePath: cliOptions.executablePath,
     headless: cliOptions.headless,
+    chromiumSandbox: !!cliOptions.sandbox,
```
pavelfeldman, line 131: You just enforced the chromium sandbox by default

tests/config.spec.ts
Skn0tt: nothing, I saw an opportunity to refactor - but it seems I don't understand what I was doing

src/config.ts
Skn0tt: good catch!

pavelfeldman (approved):

PR #479, opened 2025-05-27
chore: make library test run under older Node versions

Skn0tt:
This test only worked with `--experimental-require-module`, which is only enabled by default on recent Node versions. Changing the test to use a dynamic import tests the same regression, but works on older versions.


Skn0tt:
fixes CI failure introduced in https://github.com/microsoft/playwright-mcp/pull/476

pavelfeldman (approved):

Issue #474, opened 2025-05-25
Enhancement: Add force click tools for elements that bypass actionability checks


morganroux:
## Problem

  Sometimes elements are technically clickable but Playwright's standard actionability checks prevent interaction when elements are:
  - Temporarily covered by overlays or animations
  - In transitional states during UI changes
  - Behind other elements but still accessible

  The current browser_click and browser_screen_click tools fail in these scenarios, leaving no fallback option for stubborn elements.

  ## Proposed Solution

  Add force click variants that bypass actionability checks:
  - browser_click_mouse_force for snapshot mode (using direct mouse actions)
  - browser_screen_click_force for vision mode (using coordinate-based force clicking)

  ## Implementation

  I have already implemented this solution and can open a PR if the enhancement would be welcome. The implementation maintains consistency with existing tool patterns and includes proper error handling.

pavelfeldman:
What are you bypassing?

pavelfeldman:
Looks like the snapshot is wrong. We click the first occurrence and we hit the image nested in the second.

<img width="555" alt="Image" src="https://github.com/user-attachments/assets/3fb4eba1-08a9-4694-9009-c2081d1aa357" />

dpost:
+1 for this. It'd also be helpful to have some sort of workaround (maybe browser_execute_javascript?) for tricky automations

pavelfeldman:
Closing as non-actionable, but with the new `browser_evaluate` command you can invoke `click()`

PR #489, opened 2025-05-28
feat: add force click tool for snapshot mode

morganroux:
This PR solves the issue https://github.com/microsoft/playwright-mcp/issues/474
This "force click" method bypasses actionability checks for stubborn elements.

- Add browser_click_mouse_force for snapshot mode using direct mouse actions


morganroux:

pavelfeldman:
I don't think that's the way we should solve it. Your original click should work - I'll take a look at it upstream.

PR #495, opened 2025-05-30
chore: reuse browser in server mode

pavelfeldman:


src/browserContextFactory.ts
```diff
@@ -0,0 +1,202 @@
...
+          throw new Error(`Browser specified in your config is not installed. Either install it (likely) or change the config.`);
+        if (error.message.includes('ProcessSingleton') || error.message.includes('Invalid URL')) {
+          // User data directory is already in use, try again.
+          await new Promise(resolve => setTimeout(resolve, 1000));
+          continue;
+        }
+        throw error;
+      }
+    }
+    throw new Error(`Browser is already in use for ${userDataDir}, use --isolated to run multiple instances of the same browser`);
```
canoziia, line 177: Hello, what does this line mean? After upgrading to v0.0.28, I cannot use --user-data-dir anymore

mxschmitt: Could you file a new issue? Could you try to set the `DEBUG=pw:browser` env var? This might yield more debug output of why its not able to launch.

canoziia: After set `DEBUG=pw:browser` I found that's because my container hostname was always changing so chrome won't start. Thank you.

PR #504, opened 2025-05-31
POC: setup CORs headers

ezyang:
This PR can't be merged as is, but hopefully it is illustrative.  Currently, it's
not possible to directly talk to playwright-mcp server over SSE directly from
claude.ai in the browser with claude-mcp extension because there aren't
relaxed enough CORS headers to allow the browser extension to query the MCP
server.  It would definitely be helpful if this is configurable.  This PR
empirically works to unblock me for this case, but clearly it cannot be merged
as is.


pavelfeldman:
https://github.com/microsoft/playwright-mcp/pull/450/files is trying to address it

PR #526, opened 2025-06-05
fix: improve MCP server resilience and error handling

gannonh:
## Summary
- Adds 30-second timeout wrapper for browser actions to prevent indefinite hangs
- Implements graceful error handling for browser disconnections and unresponsive pages  
- Adds global process error handlers to prevent server crashes from uncaught exceptions
- Enhances error reporting in tool responses while maintaining stdio transport compatibility
- Safely handles page information retrieval and snapshot capture failures
- Restores missing URL import that was causing "Invalid URL: undefined" errors
- Adds packageJSON export for proper module resolution

## Problem Solved
Previously, when browser interactions encountered problematic web applications (e.g., those with infinite render loops), the browser process would become unresponsive and cause the entire MCP server to crash with "transport closed" errors. This made the server unusable with certain websites.

## Test Plan
- [x] Tested against problematic chat application that previously caused crashes
- [x] Verified MCP server remains responsive after browser errors  
- [x] Confirmed error messages are properly reported to users
- [x] Validated that stdio transport compatibility is maintained
- [x] Checked that both manual and automated interactions work correctly

## Technical Details
The changes focus on preventing process termination while maintaining full functionality. Error handling is designed to be transparent to users while providing helpful feedback when browser issues occur.

Key improvements:
- Browser action timeout prevents infinite waits
- Graceful degradation when browser becomes unresponsive
- Process-level error handlers prevent crashes
- Enhanced error reporting without breaking stdio transport

## Backwards Compatibility
All changes are backwards compatible and maintain existing API behavior while improving reliability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ playwright-report/
 
 .idea
 .DS_Store
+context/
+logs
+CLAUDE.md
+.mcp.json
--- a/src/context.ts
+++ b/src/context.ts
@@ -15,6 +15,7 @@
  */
 
 import fs from 'node:fs';
+import url from 'node:url';
 import os from 'node:os';
 import path from 'node:path';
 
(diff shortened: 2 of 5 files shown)


gannonh:

pavelfeldman:
It is always best to start with filing an issue.

PR #547, opened 2025-06-13
feat: add video recording option

raman-verto:
This PR exposes video recording cli options in playwright
Ref: https://playwright.dev/docs/videos


raman-verto:

PR #548, opened 2025-06-13
chore(extension): start relay before creating MCP server

yury-s:
* HTTPS server launched and the relay server is created before MCP server. This way we can pass CDP endpoint to its constructor.
* MCP HTTP transport is added to precreated HTTP server.
* A bunch of renames to fix style issues.

### DIFF
--- a/src/transport.ts
+++ b/src/transport.ts
@@ -100,11 +100,13 @@ async function handleStreamable(server: Server, req: http.IncomingMessage, res: 
 export async function startHttpServer(config: { host?: string, port?: number }): Promise<http.Server> {
   const { host, port } = config;
   const httpServer = http.createServer();
-  // eslint-disable-next-line no-console
-  const errorListener = (error: Error) => console.error('Error starting HTTP server', error);
-  httpServer.on('error', errorListener);
-  await new Promise<void>(resolve => httpServer.listen(port, host, resolve));
-  httpServer.removeListener('error', errorListener);
+  await new Promise<void>((resolve, reject) => {
+    httpServer.on('error', reject);
+    httpServer.listen(port, host, () => {
+      resolve();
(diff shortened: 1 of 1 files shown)


src/transport.ts
```diff
@@ -97,18 +97,27 @@ async function handleStreamable(server: Server, req: http.IncomingMessage, res:
   res.end('Invalid request');
 }
 
-export async function startHttpTransport(server: Server): Promise<http.Server> {
+export async function startHttpServer(config: { host?: string, port?: number }): Promise<http.Server> {
+  const { host, port } = config;
+  const httpServer = http.createServer();
+  // eslint-disable-next-line no-console
+  const errorListener = (error: Error) => console.error('Error starting HTTP server', error);
+  httpServer.on('error', errorListener);
```
mxschmitt: I think we should reject the promise in line 106 instead of a stderr message.

src/transport.ts
```diff
@@ -97,18 +97,27 @@ async function handleStreamable(server: Server, req: http.IncomingMessage, res:
...
+  await new Promise<void>(resolve => httpServer.listen(port, host, resolve));
+  httpServer.removeListener('error', errorListener);
+  return httpServer;
+}
+
+export function startHttpTransport(httpServer: http.Server, mcpServer: Server) {
```
mxschmitt, line 113:
```suggestion
export function bindServerToTransport(httpServer: http.Server, mcpServer: Server) {
```

maybe?

mxschmitt (approved):

src/transport.ts
yury-s: I'd leave it as is, it reflects the difference with stdio transport in the next line. Maybe there is a better name, but I feel like bindServerToTransport doesn't improve the readability.

PR #554, opened 2025-06-16
chore: connect to browser server

Skn0tt:
Counterpart of https://github.com/microsoft/playwright/pull/36327.


Skn0tt:
revision 2: https://github.com/microsoft/playwright/pull/36382

PR #561, opened 2025-06-17
chore: improve "ref not found" error message

Skn0tt:
Helps the model better understand the error cause.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -134,7 +134,11 @@ export class Context {
 
   async run(tool: Tool, params: Record<string, unknown> | undefined) {
     // Tab management is done outside of the action() call.
-    const toolResult = await tool.handle(this, tool.schema.inputSchema.parse(params || {}));
+    const toolResult = await tool.handle(this, tool.schema.inputSchema.parse(params || {})).catch(e => {
+      if (e instanceof Error && /locator._generateLocatorString: Timeout .* exceeded/.test(e.message))
+        throw new Error('Ref not found. Make sure the element is present on the page before using it in the tool.');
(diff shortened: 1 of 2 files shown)


src/context.ts
```diff
@@ -134,7 +134,11 @@ export class Context {
 
   async run(tool: Tool, params: Record<string, unknown> | undefined) {
     // Tab management is done outside of the action() call.
-    const toolResult = await tool.handle(this, tool.schema.inputSchema.parse(params || {}));
+    const toolResult = await tool.handle(this, tool.schema.inputSchema.parse(params || {})).catch(e => {
+      if (e instanceof Error && /locator._generateLocatorString: Timeout .* exceeded/.test(e.message))
```
dgozman: Shouldn't this be in `generateLocator()` helper instead?

src/context.ts
```diff
@@ -134,7 +134,11 @@ export class Context {
...
+        throw new Error('Ref not found. Make sure the element is present on the page before using it in the tool.');
```
dgozman: Some suggestion for inspiration, to make it more action-oriented.

```suggestion
        throw new Error('Element with this ref has been remove. Take a new snapshot to see the elements currently present on the page.');
```

src/context.ts
Skn0tt: ah, yes. that makes total sense.

src/context.ts
Skn0tt: Good idea! I updated the wording to a mixture of the two.

dgozman (approved):

Issue #571, opened 2025-06-19
allow to configure locale and timezoneId when starting mcp server


sabbadino-ca:
How can i inject some thing like this in the mcp server playwright.config.ts

  use: {
        ..
        locale: 'it-IT',
        timezoneId: 'Europe/Rome',
    },

It seems to me that is not supported in the Configuration file

I am having problem with date inputs 

thanks

Skn0tt:
You can put both under the `browser.contextOptions` config field.

sabbadino-ca:
Hi, I ma having problem having it working 

"Command": "npx",
"Arguments": [ "@playwright/mcp@latest", "--isolated", "--config", "./Templates/playwright-settings.json" ],

settings file 

```
{
    "contextOptions": {
        "locale": "it-IT",
         "viewport": { "width": 2500, "height": 1400}
    }
}
```
still this setting are ignored .. 

What am i doing wrong ? I can i debug the issue ? Do you have a working sample ?
thanks 
Enrico

Skn0tt:
It needs to be `browser.contextOptions`, not `contextOptions`:
```diff
{
+ "browser": {
      "contextOptions": {
          "locale": "it-IT",
           "viewport": { "width": 2500, "height": 1400}
      }
+ }
}
```

sabbadino-ca:
how stupid i am, with you suggestion it does work :) 
thanks all

Skn0tt:
Not stupid, just confused because we're just lacking proper tooling to make this intuitive :D Your issue gave me the idea to https://github.com/microsoft/playwright-mcp/pull/574, that should prevent confusion like this.

Issue #546, opened 2025-06-13
Ability to record video


RobertBroersma:
We would love to have the video recording capabilities of playwright in the MCP. 

A potential use-case:
If we have a long list of instructions this could automatically generate a nice tutorial video for our users that's always up-to-date.

Skn0tt:
Hi Robert! You can pass [`recordVideo`](https://playwright.dev/docs/api/class-browser#browser-new-context) into the `contextOptions` via `--config`: https://github.com/microsoft/playwright-mcp/blob/main/config.d.ts

Let me know how this works for you.

RobertBroersma:
Hi @Skn0tt thanks! I tried this out, but it caused the MCP server to silently not start. I'll try again and see if I can get some kind of error message.

RobertBroersma:
I can't seem to get a good error message out of it. Whenever I add the video recording config and try to visit a url, it quickly opens a browser window, which immediately closes/crashes and it says

```
Error: Browser specified in your config is not installed. Either install it (likely) or change the config.
```

But it does work without the recording config.

This also happens on the playwright-mcp-world one.

I'm on macOS Sonoma 14.5

Skn0tt:
It works fine for me:

```json
{
  "browser": {
    "browserName": "chromium",
    "launchOptions": {
      "channel": "msedge"
    },
    "contextOptions": {
      "recordVideo": {
        "dir": "/Users/skn0tt/dev/microsoft/playwright-mcp/video"
      }
    }
  }
}
```

Prompt: `go to playwright.dev, click around, then close the browser`

Resulting video:

https://github.com/user-attachments/assets/47479934-6692-4b2e-bbf5-b1a0101a671d

RobertBroersma:
Yeah, strange. It's likely something on my side. If I figure out what I'll circle back. Thanks!

jflemingframe:
happens for me as well
trying chromium/chrome

traces seems to be a decent alternative but I'd prefer video.

ithiria894:
Playwright's \`recordVideo\` config works great when you can get it set up — but for those hitting issues (like the silent crash above), or who just want a simpler workflow, I built **[Pagecast](https://github.com/mcpware/pagecast)**.

It wraps Playwright's recording into MCP tools so you don't need config files:

- **\`record_and_gif\`** — one tool call: open URL → record → optimized GIF (two-pass ffmpeg palette, no manual conversion)
[...]

PR #574, opened 2025-06-19
chore: add json schema file

Skn0tt:
We're seeing a multiple issues (https://github.com/microsoft/playwright-mcp/issues/571, https://github.com/microsoft/playwright-mcp/issues/546) about discovery for `config.json` properties. This PR gives the config file Intellisense, by generating a json-schema file and recommending it in our README.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ playwright-report/
 
 .idea
 .DS_Store
+config.json
--- a/README.md
+++ b/README.md
@@ -253,6 +253,23 @@ using the `--config` command line option:
 npx @playwright/mcp@latest --config path/to/config.json
 ```
 
+```json
+// path/to/config.json
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/playwright-mcp/main/config.schema.json",
+  "browser": {
+    "browserName": "chromium",
+    "launchOptions": {
+      "channel": "msedge"
+    },
+    "contextOptions": {
+      "timezoneId": "Europe/Berlin",
+      "locale": "de-DE"
+    }
+  }
+}
+```
+
 <details>
 <summary>Configuration file schema</summary>
 
--- a/config.d.ts
+++ b/config.d.ts
@@ -18,6 +18,32 @@ import type * as playwright from 'playwright';
 
 export type ToolCapability = 'core' | 'tabs' | 'pdf' | 'history' | 'wait' | 'files' | 'install' | 'testing';
 
+type LaunchOptions = Omit<
+  playwright.LaunchOptions,
+    'handleSIGHUP'
+  | 'handleSIGINT'
+  | 'handleSIGTERM'
+  | 'logger'
+>;
+
+type ClientCertificate = Omit<
+  NonNullable<playwright.BrowserContextOptions['clientCertificates']>[number],
+    'cert' | 'key' | 'pfx'
+>;
+
+interface RecordHar extends Omit<NonNullable<playwright.BrowserContextOptions['recordHar']>, 'urlFilter'> {
+  urlFilter?: string;
+}
+
+interface BrowserContextOptions extends Omit<
+  playwright.BrowserContextOptions,
+    'logger'
+  | 'clientCertificates'
+> {
+  clientCertificates?: ClientCertificate[];
+  recordHar?: RecordHar;
+}
+
 export type Config = {
   /**
    * The browser to use.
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ playwright-report/
 
 .idea
 .DS_Store
+config.json
```
mxschmitt:
```suggestion
/config.json
```

README.md
```diff
@@ -253,6 +253,23 @@ using the `--config` command line option:
 npx @playwright/mcp@latest --config path/to/config.json
 ```
 
+```json
```
mxschmitt:
```suggestion
```jsonc
```

This doesn't mark the line 257 as red.

config.d.ts
```diff
@@ -18,6 +18,32 @@ import type * as playwright from 'playwright';
...
+type ClientCertificate = Omit<
+  NonNullable<playwright.BrowserContextOptions['clientCertificates']>[number],
+    'cert' | 'key' | 'pfx'
+>;
+
+interface RecordHar extends Omit<NonNullable<playwright.BrowserContextOptions['recordHar']>, 'urlFilter'> {
+  urlFilter?: string;
+}
+
+interface BrowserContextOptions extends Omit<
```
mxschmitt: Why doing the clientCertificates/recordHar stuff?

mxschmitt (approved):

config.d.ts
Skn0tt: Without it, the output starts containing a json-schema version of Node.js' `Buffer`, which is impossible to provide via JSON.

```suggestion
// remove all non-JSON-compatible fields, like Buffers or Logger objects
interface BrowserContextOptions extends Omit<
```

.gitignore
Skn0tt: what does the double slash do? I only know a single slash.

mxschmitt: Should be just `/config.json` - had a typo. So it only takes the root config.json. Maybe we need config.json in the future for tests so that protects us.

config.d.ts
```diff
@@ -18,6 +18,32 @@ import type * as playwright from 'playwright';
...
 
+type LaunchOptions = Omit<
+  playwright.LaunchOptions,
+    'handleSIGHUP'
+  | 'handleSIGINT'
+  | 'handleSIGTERM'
+  | 'logger'
+>;
+
+type ClientCertificate = Omit<
```
mxschmitt: I'm not sure if we want to do this. In theory there are two kinds of configs:
- a) config which you pass to createConnection(config)
- b) config which you pass via --config

a) does support Buffers while b) does not.

We should probably split up the config types.

Skn0tt: Good point, yes. Let me split it up.

Skn0tt: done in https://github.com/microsoft/playwright-mcp/pull/574/commits/75a1f5bfa8a6713e32f9b0a6f15c45192439e824

README.md
```diff
@@ -253,6 +253,23 @@ using the `--config` command line option:
 npx @playwright/mcp@latest --config path/to/config.json
 ```
 
+```jsonc
```
pavelfeldman, line 256: Let's explicitly say that this is an escape hatch that enables all power of Playwright configurability, but might break MCP functionality, or go away in the future.

MCP is a high-level API, so allowing low-level configurability for it does not allow us to elevate the level of abstraction for the browsing, we get stuck with the low levelness of the Playwright itself.

Skn0tt: done in https://github.com/microsoft/playwright-mcp/pull/574/commits/f58418227d860c7446ffc50ac1ae308f1c9faf0f

config.d.ts
```diff
@@ -18,7 +18,37 @@ import type * as playwright from 'playwright';
...
+  interface RecordHar extends Omit<NonNullable<playwright.BrowserContextOptions['recordHar']>, 'urlFilter'> {
+    urlFilter?: string;
+  }
+
+  interface BrowserContextOptions extends Omit<
+    playwright.BrowserContextOptions,
+      'logger'
+    | 'clientCertificates'
+  > {
+    clientCertificates?: ClientCertificate[];
```
pavelfeldman, line 47: Anyone who reads this (including myself) will be wondering why this type is patched. Could you elaborate in a comment?

Skn0tt: done in https://github.com/microsoft/playwright-mcp/pull/574/commits/e3e9793c17a854ab154f3d5690e8c75718c077cb

PR #575, opened 2025-06-19
Add instructions to install in Goose

joahg:
[Goose](https://block.github.io/goose/) is a popular MCP client distributed by Block. This adds instructions to use playwright-mcp in Goose to the README


joahg:
GitHub markdown doesn't support custom protocols 😭

PR #582, opened 2025-06-20
vscode impl

jwangxx:


mxschmitt:
I don't think we want to have it in Go, esp. since Go is not an official Playwright language binding. Please start opening an issue first before you create PRs, thanks!

PR #590, opened 2025-06-22
Explicitly declare the dependencies

quangIO:
By default, npm hoist undeclared dependencies to node_modules. However, this behavior doesn't apply to pnpm. This allows building with pnpm without [shamefully-hoist](https://pnpm.io/settings#shamefullyhoist).


pavelfeldman:
- We don't use pnpm and it is not covered in the tests, so it will regress.
- Different versions of playwright and playwright-core in deps is a recipe for disaster!

PR #591, opened 2025-06-24
chore: basic client loop for integration testing

pavelfeldman:


src/client/loop.ts
```diff
@@ -0,0 +1,135 @@
...
+  type: 'function',
+  function: {
+    name: 'done',
+    description: 'Call this tool to indicate that the task is complete.',
+    parameters: {},
+  },
+};
+
+async function connectToMCP(): Promise<{ client: Client, tools: Tool[] }> {
+  const transport = new StdioClientTransport({
```
Skn0tt, line 42: If you want to use it for testing, maybe an in-memory transport is better because it allows easier debugging.
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/inMemory.ts

src/client/loop.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  for (const step of steps)
+    await performStep(step, client, openai, openAITools);
+
+  await client.close();
+}
+
+async function performStep(step: string, client: Client, openai: OpenAI, tools: OpenAI.Chat.Completions.ChatCompletionTool[]) {
+  console.log('Perform step:', step);
+  const messages: OpenAI.Chat.Completions.ChatCompletionMessageParam[] = [
```
Skn0tt, line 82: I'm surprised that chat history is discarded between steps. Keeping history between the steps would be closer to how our users use the tool. What's the intention behind it?

Skn0tt (approved):

PR #592, opened 2025-06-24
feat: assert_visible tool

dgozman:


tests/assert.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+- Page URL: ${server.PREFIX}
+- Page Title: Title
+- Page Snapshot
+\`\`\`yaml
+- button "Submit" [ref=e2]
+\`\`\`
+`);
+
+  // Wait for the button to become hidden
+  await new Promise(f => setTimeout(f, 3000));
```
Skn0tt, line 81: can we lower these to 100ms/200ms so we don't slow down unnecessarily?

tests/assert.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  expect(await client.callTool({
+    name: 'browser_assert_visible',
+    arguments: {
+      element: 'Submit button',
+      ref: 'e2',
+    },
+  })).toContainTextContent(`
+Error: Expected Submit button to be visible, but got "hidden".
```
Skn0tt, line 90: I have a hunch that this response should also contain the generated code. I'll play around with that after landing.

tests/fixtures.ts
```diff
@@ -147,6 +148,8 @@ export const test = baseTest.extend<TestFixtures & TestOptions, WorkerFixtures>(
 
   mcpMode: [undefined, { option: true }],
 
+  mcpIncludeAsserts: [false, { option: true }],
```
Skn0tt: I understand we want to flag this in prod, but I don't see a problem with enabling it unconditionally in our test suite.

src/tools/assert.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const elementSchema = z.object({
+  element: z.string().describe('Human-readable element description used to obtain permission to interact with the element'),
+  ref: z.string().describe('Exact target element reference from the page snapshot'),
+});
+
+const assertVisible = defineTool({
+  capability: 'testing',
+  schema: {
+    name: 'browser_assert_visible',
```
Skn0tt, line 24: After landing, i'll play around with namespacing the test-specific tools with `playwright_` instead of `browser_`.

src/tools/assert.ts
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { z } from 'zod';
+import { defineTool } from './tool.js';
+import { generateLocator } from './utils.js';
+
+const elementSchema = z.object({
```
Skn0tt: nit: import this from `snapshot.ts`

src/tools/assert.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    return {
+      code,
+      action: async () => {
+        const result = await (locator as any)._expect('to.be.visible', { timeout: 5000 });
+        if (!result.matches) {
+          const error = [`Expected ${params.element} to be visible, but got "${result.received}".`];
+          if (result.log?.length)
+            error.push(``, `Log:`, ...result.log);
+          throw new Error(error.join('\n'));
```
Skn0tt, line 48: Throwing an error here makes us return `isError: true`, which the spec mentions is for ["API failures, Invalid input data, Business logic errors"](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling). I'm not sure if a failing assertion is a business logic errors. Did you play around with returning some content instead?

src/tools/assert.ts
```diff
@@ -0,0 +1,64 @@
...
+        }
+      },
+      captureSnapshot: true,
```
Skn0tt: An assertion shouldn't change anything about the page contents, let's not capture a new snapshot.
```suggestion
      captureSnapshot: false,
```

src/tools/assert.ts
dgozman: I want this to be a failure, because that's easier to see for the user. That said, triggering this is very hard, because LLM picks visible elements 😄

tests/assert.spec.ts
dgozman: Not really.

src/tools/assert.ts
Skn0tt: Makes sense! `isError` might be shown differently in the client, that's a good reason.

Skn0tt:
You need to update the tool list test.

examples/generate-test.md
```diff
@@ -1,10 +1,20 @@
-Use Playwright tools to generate test for scenario:
+## Instructions
```
pavelfeldman, line 1: You should only include instructions here if you intend to remove the generate_test tool.

examples/generate-test.md
```diff
@@ -1,10 +1,20 @@
...
+
+## Steps
 
 1. Open the [Microsoft Playwright GitHub repository](https://github.com/microsoft/playwright).
-2. Click on the **Pull requests** tab.
-3. Find and open the pull request titled **"chore: make noWaitAfter a default"**.
-4. Switch to the **Checks** tab for that pull request.
-5. Expand the **infra** check suite to view its jobs.
-6. Click on the **docs & lint** job to view its details.
+2. Switch to the "Pull requests" tab.
```
pavelfeldman, line 16: This script was generated by AI, I'd rather keep it as is (* instead of ")

src/tools.ts
```diff
@@ -32,7 +33,10 @@ import wait from './tools/wait.js';
 
 import type { Tool } from './tools/tool.js';
 
+const assertTools = process.env.PWMCP_INCLUDE_ASSERTS ? assert : [];
```
pavelfeldman, line 36: CLI is supposed to controls it, no need for env. You can add 'default' alias that includes default categories for convenience.

src/tools/assert.ts
pavelfeldman: > After landing, i'll play around with namespacing the test-specific tools with playwright_ instead of browser_.

Let's not.

src/tools/assert.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const code = [
+      `// Assert ${params.element} is visible`,
+      `await expect(page.${await generateLocator(locator)}).toBeVisible();`
+    ];
+
+    return {
+      code,
+      action: async () => {
+        const result = await (locator as any)._expect('to.be.visible', { timeout: 5000 });
```
pavelfeldman, line 43: I don't think you need the actual check here as your snapshot will only include visible elements. For other assertions, let's use library (non-expect) apis to ensure the conditions as they are never "awaited". You use refs in the mcp mode, so either that or you need to reconsider your approach.

src/tools/assert.ts
```diff
@@ -0,0 +1,59 @@
...
+import { generateLocator } from './utils.js';
+import { elementSchema } from './snapshot.js';
+
+const assertVisible = defineTool({
+  capability: 'testing',
+  schema: {
+    name: 'browser_assert_visible',
+    title: 'Assert Visible',
+    description: 'Assert that an element is visible on the page',
+    inputSchema: elementSchema,
```
pavelfeldman, line 27: TLDR: You are not allowed to use element refs in assertions.

Issue #595, opened 2025-06-25
Bug: dialog blocks snapshot


route250:
## Description

When using Playwright MCP, if a browser_click action triggers a native browser dialog (alert, confirm, prompt), the call currently hangs until the dialog is handled. There is no way for the client to know that a dialog appeared, nor to retrieve its type or message for conditional handling.

## Proposal

- When browser_click triggers a dialog, the response should immediately indicate:
  - That a dialog appeared
  - The dialog type (alert, confirm, prompt)
  - The dialog message text
- This would allow clients to implement conditional logic based on the dialog content, and avoid indefinite hangs.

## Steps to Reproduce

1. Use browser_click to click a button that triggers a native dialog.
2. Observe that the call hangs and no dialog information is returned.

## Expected Behavior

- browser_click should return a response indicating a dialog appeared, with its type and message, instead of hanging.

## Actual Behavior

- browser_click hangs until the dialog is handled, and no dialog information is available to the client.

## Environment

- Playwright MCP version:0.0.29

## Additional Context

This enhancement would make it possible to automate workflows that require conditional dialog handling, and improve robustness of MCP-based browser automation.

Skn0tt:
We have support for this with the `browser_handle_dialog` tool. If it doesn't work for you, that sounds like a bug. Could you elaborate on the behaviour you're seeing, ideally with screenshots attached, and provide us with a prompt we can run to reproduce the bug you're seeing?

route250:
I wasn’t familiar with browser_handle_dialog, so I restructured my test code to adopt it. The test script is mcp_client_test.py.

The actual application I’m trying to automate is a React-based web app. Since it’s too complex for isolated testing, I created a simplified version of the server-side code named uitest_serverA.py. When I run the test with this setup, it hangs after clicking the “Login” button.

To narrow things down further, I created an even simpler version called uitest_serverB.py. With this one, the click action does not hang — even if I remove browser_handle_dialog, the test runs through without blocking.

If this issue can be resolved, it would be really great. What do you think?

[mcp_client_test.py.gz](https://github.com/user-attachments/files/20915031/mcp_client_test.py.gz)

[uitest_serverA.py.gz](https://github.com/user-attachments/files/20915035/uitest_serverA.py.gz)

[uitest_serverB.py.gz](https://github.com/user-attachments/files/20915036/uitest_serverB.py.gz)

[...]

PR #598, opened 2025-06-26
chore: race snapshot against dialog

Skn0tt:
A dialog that pops up while we're taking a snapshot, or while `context.waitForTimeout` is running, blocks everything, leading to https://github.com/microsoft/playwright-mcp/issues/595.
We fix that by wrapping more of our code with `_raceAgainstModalDialogs`.

Closes https://github.com/microsoft/playwright-mcp/issues/595

### DIFF
--- a/tests/dialogs.spec.ts
+++ b/tests/dialogs.spec.ts
@@ -210,3 +210,31 @@ test('prompt dialog', async ({ client, server }) => {
 - generic [ref=e1]: Answer
 \`\`\``);
 });
+
+
+test('alert dialog repro', { annotation: { type: 'issue', description: 'https://github.com/microsoft/playwright-mcp/issues/595' } }, async ({ client, server }) => {
+  server.setContent('/', `
+    <form>
+        <button id="button" type="submit">Login</button>
+    </form>
+    <script>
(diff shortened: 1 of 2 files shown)


tests/dialogs.spec.ts
```diff
@@ -210,3 +210,31 @@ test('prompt dialog', async ({ client, server }) => {
...
+    name: 'browser_navigate',
+    arguments: { url: server.PREFIX },
+  })).toContainTextContent('- button "Login" [ref=e3]');
+  expect(await client.callTool({
+    name: 'browser_click',
+    arguments: {
+      element: 'Login',
+      ref: 'e3',
+    },
+  })).toContainTextContent(`"alert" dialog`);
```
pavelfeldman, line 239: Unclear why this test would work. Click will return immediately, snapshot will be empty. Then the dialog will appear.

Skn0tt: It passes because of our 1s wait: https://github.com/microsoft/playwright-mcp/blob/137b74750c5dc00305dc8bce0721c60a107fe04e/src/tools/utils.ts#L65

Skn0tt:
Closing as stale.

PR #603, opened 2025-06-26
chore(extension): convert to typescript

yury-s:


extension/src/popup.ts
```diff
@@ -14,54 +14,65 @@
...
+  private statusContainer: HTMLElement | null;
+  private actionContainer: HTMLElement | null;
+
   constructor() {
     this.currentTab = null;
-    this.bridgeUrlInput = /** @type {HTMLInputElement} */ (document.getElementById('bridge-url'));
-    this.connectBtn = /** @type {HTMLButtonElement} */ (document.getElementById('connect-btn'));
-    this.statusContainer = /** @type {HTMLElement} */ (document.getElementById('status-container'));
-    this.actionContainer = /** @type {HTMLElement} */ (document.getElementById('action-container'));
+    this.bridgeUrlInput = document.getElementById('bridge-url') as HTMLInputElement | null;
```
mxschmitt: Why can all of these be null? In theory they should be never null.

mxschmitt (approved):

Issue #420, opened 2025-05-14
[Question] Enable vision capabilities as well as snapshot capabilites


gavyn-ezell:
Is there a way to enable vision capabilities but still include default snapshot capabilities. My use case needs the ability to prioritize snapshots, but use vision when necessary.

Pbonmars-20031006:
How did you enable it ?

wernerthiago:
I implemented a multiclient mcp that does this logic, so, if the tool already exists from default server, skip the tool, if not, add the tool to the list. It should work.

samschurter:
I would like the ability to use tools from both the vision and snapshot mode as well. My use case is working with an app which embeds Google Maps. The map is drawn to a canvas, so there's no way to click a specific element and drag it to another element in order to perform map actions. However, the map is only a part of the application. We have a lot of other elements around the map as well, and interacting with those is fast and reliable with the snapshot API. I'd like to be able to use the snapshot API for most interactions, but then instruct the agent to take a screenshot and use screen-based move/click/drag when necessary.

brnaba-aws:
anyone has an implementation to share?
Thanks

wernerthiago:
> anyone has an implementation to share? Thanks

Hey @brnaba-aws ! In my case I had to give up mixing both servers to access both features. I needed to create a patch to add (in my case) a "browser click with coordinates" tool in the default mode. I can share the patch, but it is not a good practice patching code like that. So, the best approach at the of the day is to implement it as feature.

brnaba-aws:
I cloned the repo and created an Hybrid version which combines the tools from vision and Snapshot. I'll create a patch

brnaba-aws:
See #611

pavelfeldman:
Screenshots are available in the snapshot mode, do you need to click by coordinates sometimes?

brnaba-aws:
Yes, in case my case, the pdf file was opening in pdf viewer. The only way to click on download is with X,Y coordinates.

PR #611, opened 2025-06-30
Added Hybrid Mode that combines Snapshot and Vision modes

brnaba-aws:
In some cases we could have the option to use vision mode as well as snapshot mode for robustness.
This PR adds a `hybrid` flag to combine both options.

Tests have been updated.

See #420


brnaba-aws:

pavelfeldman:
It is always best to get feedback in the issue first.

PR #628, opened 2025-07-03
chore: support custom testIdAttribute

mxschmitt:
This will allow users to set a custom testId which is helpful for projects relying on a different one. Especially with testing use-cases where generation from existing code/text/logic is happening.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,9 @@ npx @playwright/mcp@latest --config path/to/config.json
 
     // Remote Playwright server endpoint
     remoteEndpoint?: string;
+
+    // testIdAttribute to use. Defaults to "data-testid"
+    testIdAttribute?: string;
   },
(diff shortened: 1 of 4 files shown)


pavelfeldman:
How would LLM know the test id to generate?

mxschmitt:
> How would LLM know the test id to generate?

This changes the selector we return on e.g. a click by a ref. Previously if the mycorp.com has used `data-testid` we already returned `getByTestId`. With this PR custom `testIdAttributes` are supported and are returned as `getByTestId` as well. So the LLM just uses what the MCP server (click) previously returned.

pavelfeldman:
I see what you are saying. You should plumb it in the generateLocatorString then.

pavelfeldman:
Closing as stale

PR #662, opened 2025-07-14
docs: Adds instructions on how to install with Zencoder

viborc:
This PR improves the docs by providing detailed instructions on how to install Playwright MCP in [Zencoder](https://zencoder.ai), listing both direct config options and a selection from the MCP Library available in Zencoder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,27 @@ Follow the MCP install [guide](https://github.com/google-gemini/gemini-cli/blob/
 ```
 </details>
 
+<details>
+<summary><b>Install in Zencoder</b></summary>
+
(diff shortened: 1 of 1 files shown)


viborc:

pavelfeldman:
Our installation list grew large and clients converged on a pretty much the same config anyways, so we are not accepting the new ones.

viborc:
@pavelfeldman I get it - but I somewhat disagree since Zencoder's approach is also a bit different in a way that it allows for installation of Playwright from its MCP Library, not just by pasting a config, and I think I mentioned that in the actual commit. Hope this warrants a reconsideration?

In any case, thanks for the quick reply earlier!

Issue #669, opened 2025-07-15
Support page.on('pageerror') in tab.ts to get uncaught errors from console


ilyaev:
There is no way to get uncaught exceptions in loaded page even though they appear in console in browser - tool 'browser_console_messages' don't return them because they don't happens in 'console' event. Would be really useful to have opportunity to get this information too.

PR #671, opened 2025-07-15
chore: include page errors in console messages

pavelfeldman:
Fixes https://github.com/microsoft/playwright-mcp/issues/669

### DIFF
--- a/src/tab.ts
+++ b/src/tab.ts
@@ -17,14 +17,14 @@
 import * as playwright from 'playwright';
 
 import { PageSnapshot } from './pageSnapshot.js';
+import { callOnPageNoTrace } from './tools/utils.js';
 
 import type { Context } from './context.js';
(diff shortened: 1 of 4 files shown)


src/tab.ts
```diff
@@ -118,3 +119,32 @@ export class Tab {
...
+    type: message.type(),
+    text: message.text(),
+    toString: () => `[${message.type().toUpperCase()}] ${message.text()} @ ${message.location().url}:${message.location().lineNumber}`,
+  };
+}
+
+function pageErrorToConsoleMessage(errorOrValue: Error | any): ConsoleMessage {
+  if (errorOrValue instanceof Error) {
+    return {
+      type: undefined,
```
yury-s, line 140: type: 'PageError' ?

pavelfeldman: I want to print it the same way it is in console, there is no PageError prefix in console.

Issue #570, opened 2025-06-18
browser_tab_select must specify that is uses 1-based indexing


alextatarinov:
As [evidenced by tests](https://github.com/microsoft/playwright-mcp/blob/906698809823372c16a7fc9486fa90c316192507/tests/tabs.spec.ts#L87-L113) and direct testing, browser_tab_select expects 1-based indexing. 

Without it being explained in the tool definition, most LLMs will default to 0-based indexing, and they are very stubborn about it unless given very specific prompts.

I am not even sure that adjusting tool definition will help much - maybe a better route is to make the tool 0-indexed as standard for programming APIs.

Claude Code with Sonnet 3.7 example - it tries to use `index: 1` **four times** before deciding to use `index: 2`.

```
I've opened a new tab. It looks like you now have three tabs open:
  1. Google
  2. X (Twitter)
  3. Current tab (blank)

  The new tab is currently blank. Would you like to navigate to a specific website in this tab?

> switch to twitter

● I'll switch to the Twitter tab.

● playwright:browser_tab_select (MCP)(index: 1)
  ⎿  - Ran Playwright code:            
     ```js
     // <internal code to select tab 1>
     … +68 lines (ctrl+r to expand)

● Let me select the Twitter tab (tab #2).

● playwright:browser_tab_select (MCP)(index: 1)
  ⎿  - Ran Playwright code:            
     ```js
     // <internal code to select tab 1>
     … +68 lines (ctrl+r to expand)

● Let me try switching to the Twitter tab again with the correct index:
  ⎿  ### Open tabs                                    
     - 1: (current) [Google] (https://www.google.com/)
     - 2: [(9) Home / X] (https://x.com/home)

● playwright:browser_tab_select (MCP)(index: 1)
  ⎿  - Ran Playwright code:            
     ```js
     // <internal code to select tab 1>
     … +68 lines (ctrl+r to expand)

● Let me try with the correct index. The tab indices are zero-based, so Twitter would be index 1:

● playwright:browser_tab_select (MCP)(index: 1)
  ⎿  - Ran Playwright code:            
     ```js
     // <internal code to select tab 1>
     … +68 lines (ctrl+r to expand)

● It seems there might be an issue with the tab switching. Let me try one more approach using the
  tab index that matches the displayed numbers in the list:

● playwright:browser_tab_select (MCP)(index: 2)
  ⎿  - Ran Playwright code:            
     ```js
     // <internal code to select tab 2>
[...]

PR #674, opened 2025-07-16
chore: make tab indexes 0-based

pavelfeldman:
Fixes https://github.com/microsoft/playwright-mcp/issues/570

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -101,7 +101,7 @@ export class Context {
   }
 
   async selectTab(index: number) {
-    this._currentTab = this._tabs[index - 1];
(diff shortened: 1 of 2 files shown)


mxschmitt (approved):

src/context.ts
```diff
@@ -101,7 +101,7 @@ export class Context {
   }
 
   async selectTab(index: number) {
-    this._currentTab = this._tabs[index - 1];
```
yury-s, line 104: Should the tool description mention that the index is 0-based?

pavelfeldman: When you run "list", they have indexes.

PR #679, opened 2025-07-16
 chore: turn vision into capability

pavelfeldman:
Fixes https://github.com/microsoft/playwright-mcp/issues/420


src/tools/mouse.ts
```diff
@@ -121,12 +85,12 @@ const click = defineTool({
   },
 });
 
-const drag = defineTool({
-  capability: 'core',
+const mouseDrag = defineTool({
+  capability: 'vision',
   schema: {
-    name: 'browser_screen_drag',
+    name: 'browser_mouse_drag_xy',
```
yury-s, line 91: drop _xy suffix?

src/tools/wait.ts
```diff
@@ -40,7 +40,7 @@ const wait: ToolFactory = captureSnapshot => defineTool({
 
     if (params.time) {
       code.push(`await new Promise(f => setTimeout(f, ${params.time!} * 1000));`);
-      await new Promise(f => setTimeout(f, Math.min(10000, params.time! * 1000)));
+      await new Promise(f => setTimeout(f, Math.min(30000, params.time! * 1000)));
```
yury-s, line 43: is it related?

yury-s (approved):

src/tools/mouse.ts
pavelfeldman: I want LLM to be reluctant to call those

src/tools/wait.ts
pavelfeldman: Folks were complaining about it

PR #686, opened 2025-07-17
chore: pass MCP client info to BrowserContextFactory

yury-s:


pavelfeldman (commented):
Let's wait for the entire story to be ready, I think there will be many more changes. We need to d.ts them, so that's a lot of effort. Easier to make it ready and open the stable version.

PR #690, opened 2025-07-17
feat: added video record

trainfrombasics:
Added video record - some times we need to do some actions on the browser and need to save the video recording for further reference.


pavelfeldman:
It is always best to start with filing an issue.

PR #705, opened 2025-07-18
feat(docker): allow exposing remote access via port

agg23:
More advanced use cases of tools like MCP will require multiple Docker containers, particularly on hosts other than the user's machine. Expose the HTTP transport through a `REMOTE_HTTP` envvar and allow the user to connect to the server through the Docker firewall.

Docker logs this to the console as startup:
```
Listening on http://[DOCKER_HOST_IP]:8931
Put this in your client config:
{
  "mcpServers": {
    "playwright": {
      "url": "http://[DOCKER_HOST_IP]:8931/sse"
    }
  }
}
If your client supports streamable HTTP, you can use the /mcp endpoint instead.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,10 @@ FROM base
 ARG PLAYWRIGHT_BROWSERS_PATH
 ARG USERNAME=node
 ENV NODE_ENV=production
+ENV REMOTE_HTTP=
+
+# Used when REMOTE_HTTP is set
+EXPOSE 8931
 
 # Set the correct ownership for the runtime user on production `node_modules`
 RUN chown -R ${USERNAME}:${USERNAME} node_modules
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -66,4 +70,4 @@ COPY --chown=${USERNAME}:${USERNAME} cli.js package.json ./
 COPY --from=builder --chown=${USERNAME}:${USERNAME} /app/lib /app/lib
 
 # Run in headless and only with chromium (other browsers need more dependencies not included in this image)
-ENTRYPOINT ["node", "cli.js", "--headless", "--browser", "chromium", "--no-sandbox"]
+ENTRYPOINT ["sh", "-c", "PLAYWRIGHT_DOCKER=1 node cli.js --headless --browser chromium --no-sandbox ${REMOTE_HTTP:+--host 0.0.0.0 --port 8931}"]
```
mxschmitt: This seems hard to follow and seems like a good signal for an entrpoint.sh?

agg23: Fair point

pavelfeldman:
It is always best to start with filing an issue

PR #706, opened 2025-07-18
chore(docs): make VSCode match other README sections

agg23:


pavelfeldman (approved):
Note that we had those buttons on top featuring VS Code.

agg23:
> Note that we had those buttons on top featuring VS Code.

I don't like them being in both places, but I don't like the VSCode section missing information. I realize we want to highlight VSCode functionality as well.

PR #716, opened 2025-07-19
chore: use streamable http by default

pavelfeldman:


src/transport.ts
```diff
@@ -80,15 +82,22 @@ async function handleStreamable(server: Server, req: http.IncomingMessage, res:
   if (req.method === 'POST') {
     const transport = new StreamableHTTPServerTransport({
       sessionIdGenerator: () => crypto.randomUUID(),
-      onsessioninitialized: sessionId => {
-        sessions.set(sessionId, transport);
+      onsessioninitialized: async sessionId => {
```
yury-s, line 85: Does the caller await? If not, what if we receive a command for the session before it gets into the map?

src/transport.ts
```diff
@@ -80,15 +82,22 @@ async function handleStreamable(server: Server, req: http.IncomingMessage, res:
...
+        testDebug(`create http session: ${transport.sessionId}`);
+        const connection = await server.createConnection(transport);
+        sessions.set(sessionId, { transport, connection });
       }
     });
+
     transport.onclose = () => {
-      if (transport.sessionId)
-        sessions.delete(transport.sessionId);
+      const result = transport.sessionId ? sessions.get(transport.sessionId) : undefined;
```
yury-s, line 93: result -> session

PR #735, opened 2025-07-22
chore(extension): use free port

yury-s:


src/browserContextFactory.ts
```diff
@@ -217,7 +217,7 @@ async function injectCdpPort(browserConfig: FullConfig['browser']) {
     (browserConfig.launchOptions as any).cdpPort = await findFreePort();
 }
 
-async function findFreePort() {
+export async function findFreePort(): Promise<number> {
```
pavelfeldman: You don't want to use this method, bind to 0 instead.

src/extension/cdpRelay.ts
```diff
@@ -28,7 +28,7 @@ import { WebSocket, WebSocketServer } from 'ws';
 import debug from 'debug';
 import * as playwright from 'playwright';
 import { httpAddressToString, startHttpServer } from '../transport.js';
-import { BrowserContextFactory } from '../browserContextFactory.js';
+import { BrowserContextFactory, findFreePort } from '../browserContextFactory.js';
```
pavelfeldman: drive-by BrowserContextFactory should be imported as type

PR #736, opened 2025-07-22
chore(extension): propagate errors to the client

yury-s:


src/extension/cdpRelay.ts
```diff
@@ -205,89 +205,63 @@ export class CDPRelayServer {
 
   private async _handlePlaywrightMessage(message: CDPCommand): Promise<void> {
     debugLogger('← Playwright:', `${message.method} (id=${message.id})`);
-    if (!this._extensionConnection) {
-      debugLogger('Extension not connected, sending error to Playwright');
+    const { id, sessionId, method, params } = message;
+    try {
+      const result = await this._handleCDPCommand(method, params, sessionId);
```
pavelfeldman, line 210: What is session id?

yury-s: - Root sessionId is generated by the extension and filtered out from the incoming messages when CDP command is forwarded to the debugger.
- Child session ids are generated by the debugger and should be propagated along with the commands to it.

As discussed offline, the former can be handled in cdpRelay and not sent to the extension. I will make the change in a separate pr.

PR #738, opened 2025-07-22
chore: introduce response type

pavelfeldman:


src/tab.ts
```diff
@@ -226,28 +242,33 @@ export class Tab {
...
+        snapshot = await this.captureSnapshot();
       }
     });
-    return { actionResult, snapshot };
+
+    if (snapshot) {
+      response.addSnapshot(snapshot);
+    } else if (response.includeSnapshot()) {
+      // We are blocked on modal dialog.
+      response.addSnapshot(await this.captureSnapshot({ omitAriaSnapshot: true }));
```
yury-s, line 262: Let's have _raceAgainstModalDialogs return an explicit indicator that there is. a dialog and check it here for clarity?

src/tools/navigate.ts
```diff
@@ -57,18 +50,13 @@ const goBack = defineTabTool({
...
 
-    return {
-      code,
-      captureSnapshot: true,
-      waitForNetwork: false,
-    };
+    await tab.page.goBack();
+    response.addCode(`// Navigate back`);
+    response.addCode(`await page.goBack();`);
+    response.addSnapshot(await tab.captureSnapshot());
```
yury-s, line 59: Isn't it enough to mark the response setIncludeSnapshot?

pavelfeldman: only if i run tab.run()

PR #739, opened 2025-07-22
chore: follow up on tab snapshot capture

pavelfeldman:


src/tools/navigate.ts
```diff
@@ -54,9 +54,9 @@ const goBack = defineTabTool({
     response.setIncludeSnapshot();
 
     await tab.page.goBack();
+    response.setIncludeSnapshot();
```
yury-s, line 57: why call it twice?

yury-s (approved):

PR #746, opened 2025-07-24
chore: one tool experiment

pavelfeldman:


src/loop/onetool.ts
```diff
@@ -0,0 +1,84 @@
...
+
+dotenv.config();
+
+const __filename = url.fileURLToPath(import.meta.url);
+
+let innerClient: Client | undefined;
+const delegate = new OpenAIDelegate();
+
+const oneTool = defineTool({
+  capability: 'pdf',
```
yury-s: Why pdf?

src/server.ts
```diff
@@ -21,20 +21,23 @@ import { contextFactory as defaultContextFactory } from './browserContextFactory
...
 export class Server {
   readonly config: FullConfig;
   private _browserConfig: FullConfig['browser'];
   private _contextFactory: BrowserContextFactory;
+  readonly tools: Tool<any>[];
 
-  constructor(config: FullConfig, contextFactory?: BrowserContextFactory) {
+  constructor(config: FullConfig, tools: Tool<any>[], contextFactory?: BrowserContextFactory) {
     this.config = config;
+    this.tools = tools;
```
yury-s, line 34: Why not keep it inside `createMCPServer`, I don't see any usages of the field?

src/server.ts
```diff
@@ -21,20 +21,23 @@ import { contextFactory as defaultContextFactory } from './browserContextFactory
...
     this._browserConfig = config.browser;
     this._contextFactory = contextFactory ?? defaultContextFactory(this._browserConfig);
   }
 
   async createConnection(transport: Transport): Promise<void> {
-    const server = await createMCPServer(this.config, this._contextFactory);
+    const server = await createMCPServer(this.config, this.tools, this._contextFactory);
```
pavelfeldman, line 40: Here is the usage

PR #749, opened 2025-07-24
adding ignore https error method as a mcp tool

anushkasingh405:
adding ignore https error method as a mcp tool

### DIFF
--- a/examples/ignore-https-errors.md
+++ b/examples/ignore-https-errors.md
@@ -1,0 +1,60 @@
+# Ignore HTTPS Errors Example
+
+This example demonstrates how to use the `browser_ignore_https_errors` tool to configure the browser to ignore HTTPS certificate errors.
+
+## Use Case
+
+When testing applications that use self-signed certificates or have certificate issues during development, you may need to bypass HTTPS certificate validation.
+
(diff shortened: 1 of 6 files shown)


anushkasingh405:

pavelfeldman:
Please use --ignore-https-errors cli flag instead.

anushkasingh405:
we want to launch the chrome browser with these setting by cursor only, we are using mcp for 1st level sanity of projects    like Shared setup for test cases  
• Start a fresh Chrome browser in non-headless mode with ignore https certificate true.  
• After every action, immediately fetch and review the latest visible HTML to derive robust selectors for the next step.  
• Do NOT generate or suggest any code automation, do NOT provide interim step summaries, and do NOT ask for manual intervention.  
• Report only the final result or any errors once the entire test case finishes.
 
────────────────────
TEST CASE 1 – lOGIN
Steps  
1. Navigate to xyz.com
2. Click the “Log in” button and log in with  
   • email : xyz
   • password :123

anushkasingh405:
and we want to use this only for test environment , hence changing the configuration continuously looks like more of a hassle than just changing the prompt

PR #751, opened 2025-07-24
chore: extract pure mcp server helpers

pavelfeldman:


src/mcp/server.ts
```diff
@@ -0,0 +1,105 @@
...
+      return errorResult(`Tool "${request.params.name}" not found`);
+
+    try {
+      return await backend.callTool(tool, tool.inputSchema.parse(request.params.arguments || {}));
+    } catch (error) {
+      return errorResult(String(error));
+    }
+  });
+
+  if (backend.serverInitialized)
```
yury-s, line 99: Let's do it unconditionally in case the listener is added later.

PR #755, opened 2025-07-24
chore: extract loop tools into a separate folder

pavelfeldman:


src/loopTools/context.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  async runTask(task: string, oneShot: boolean = false): Promise<mcpServer.ToolResponse> {
+    const result = await runTask(this._delegate, this._client!, task, oneShot);
+    return {
+      content: [{ type: 'text', text: result }],
+    };
+  }
+
+  async close() {
+    await BrowserContext.disposeAll();
```
yury-s, line 63: There is already BrowserContext in Playwright API, maybe MCPContext?

pavelfeldman: All of these are MCPContexts, including the one this line is in...

yury-s (approved):

PR #764, opened 2025-07-25
chore: ping client and disconnect on connection termination

pavelfeldman:


src/mcp/server.ts
```diff
@@ -96,10 +98,46 @@ export function createServer(backend: ServerBackend): Server {
...
+  let lastId = 0;
+  const pingIds = new Set<number>();
+  onTransportMessage(transport, message => {
+    if ('id' in message)
+      pingIds.delete(message.id as number);
+  });
+  const heartbeatInterval = setInterval(() => {
+    const id = ++lastId;
+    pingIds.add(id);
+    transport.send({ id, method: 'ping', jsonrpc: '2.0', params: {} }).catch(() => {
```
yury-s: Just call `server.ping()`, no need to reimplement transport.

src/mcp/server.ts
```diff
@@ -96,10 +98,31 @@ export function createServer(backend: ServerBackend): Server {
...
+      server.ping(),
+      new Promise((_, reject) => setTimeout(() => reject(new Error('ping timeout')), 5000)),
+    ]).then(() => {
+      setTimeout(beat, 3000);
+    }).catch(() => {
+      void server.close();
+    });
+  };
+
+  // Allow incoming request before we start sending pings (sse events).
```
yury-s:
```suggestion
  // Wait for SEE channel to be established (a separate incoming GET request) before we start sending pings.
```

yury-s (approved):

src/mcp/server.ts
pavelfeldman: Addressed it via engaging upon request.

src/index.ts
```diff
@@ -27,7 +27,7 @@ import type { Server } from '@modelcontextprotocol/sdk/server/index.js';
 export async function createConnection(userConfig: Config = {}, contextGetter?: () => Promise<BrowserContext>): Promise<Server> {
   const config = await resolveConfig(userConfig);
   const factory = contextGetter ? new SimpleBrowserContextFactory(contextGetter) : contextFactory(config.browser);
-  return mcpServer.createServer(new BrowserServerBackend(config, factory));
+  return mcpServer.createServer(new BrowserServerBackend(config, factory), false);
```
yury-s, line 30: make it configurable?

src/mcp/server.ts
```diff
@@ -96,10 +102,30 @@ export function createServer(backend: ServerBackend): Server {
...
+  addServerListener(server, 'initialized', () => backend.serverInitialized?.(server.getClientVersion()));
+  addServerListener(server, 'close', () => backend.serverClosed?.());
   return server;
 }
+
+const startHeartbeat = (server: Server) => {
+  const beat = () => {
+    Promise.race([
+      server.ping(),
+      new Promise((_, reject) => setTimeout(() => reject(new Error('ping timeout')), 5000)),
```
yury-s, line 114: I'd be more generous with the timeout, 15s maybe.

src/mcp/server.ts
```diff
@@ -96,10 +102,30 @@ export function createServer(backend: ServerBackend): Server {
...
+    ]).then(() => {
```
yury-s, line 115: why not use `await`  for better readability?

yury-s (approved):

src/index.ts
pavelfeldman: Let's do that upon request. They can implement it on the call site if needed, same way we did.

src/mcp/server.ts
pavelfeldman: Let's land as is and see. 5 seconds for roundtrip is a lot for local setup, we don't think it is used on server for now.

src/mcp/server.ts
pavelfeldman: I'm making sure we do not create an infinite promise chain.

PR #767, opened 2025-07-25
chore: do not snapshot on fill

pavelfeldman:


src/tools/keyboard.ts
```diff
@@ -78,6 +76,7 @@ const type = defineTabTool({
       }
 
       if (params.submit) {
+        response.setIncludeSnapshot();
```
yury-s, line 80: I'd keep it when typing slowly too as it may well change the state of the page.

yury-s (approved):

PR #776, opened 2025-07-28
In Readme Programmatic Usage Fix

anuragasawa20:
As their connection create is written as    "await connection.sever.connect(transport)"   which is not correct as this is already connected so   "await connection.connect(transport);" is working


pavelfeldman (approved):

pavelfeldman:
Docker tests failed due to unrelated change, closing

PR #780, opened 2025-07-29
chore(extension): do not show chrome: tabs

yury-s:


extension/src/background.ts
```diff
@@ -117,7 +117,7 @@ class TabShareExtension {
 
   private async _getTabs(): Promise<chrome.tabs.Tab[]> {
     const tabs = await chrome.tabs.query({});
-    return tabs;
+    return tabs.filter(tab => tab.url && !['chrome:', 'edge:', 'devtools:'].some(scheme => tab.url!.startsWith(scheme)));
```
pavelfeldman, line 120: Maybe only allow http(s) instead?

pavelfeldman (approved):

extension/src/background.ts
yury-s: I'd like to allow navigating in the same tab too (which is chrome-extension://).

Issue #790, opened 2025-07-29
Incorrect hostnames introduced by #751


brianredbeard:
PR #751 introduced invalid hostnames which are sent to the output, which creates issues with troubleshooting:

https://github.com/microsoft/playwright-mcp/blob/65d99fe5957f368387d0f61362338917b2ce3391/src/httpServer.ts#L41-L42

When a server is configured to listen on 0.0.0.0 (for IPv4) or [::] (for IPv6), it binds to all available network interfaces on the host machine. This means it can accept connections originating from the local machine as well as from external machines on the same network.

In contrast, localhost is a hostname that resolves to the loopback interface (typically 127.0.0.1 for IPv4). A server listening specifically on localhost will only accept connections from the same machine.

Therefore, localhost is not the same as the IPv4 (0.0.0.0) or IPv6 ([::]) "listen all" addresses, because localhost implies a more restricted, local-only accessibility.  

This is especially annoying because when the hostname `localhost` _is_ supplied, the hosts resolver converts it to the IPv6 address of localhost, `[::1]`.

```
node@systemd-mcp-servers:/app$ /app/cli.js --host 0.0.0.0 --port 1234 
Listening on http://localhost:1234
Put this in your client config:
{
  "mcpServers": {
    "playwright": {
      "url": "http://localhost:1234/mcp"
    }
  }
}
For legacy SSE transport support, you can use the /sse endpoint instead.
^Cnode@systemd-mcp-servers:/app$ /app/cli.js --host 127.0.0.1 --port 1234 
Listening on http://127.0.0.1:1234
Put this in your client config:
{
  "mcpServers": {
    "playwright": {
      "url": "http://127.0.0.1:1234/mcp"
    }
  }
}
For legacy SSE transport support, you can use the /sse endpoint instead.
^Cnode@systemd-mcp-servers:/app$ /app/cli.js --host localhost --port 1234 
Listening on http://[::1]:1234
Put this in your client config:
{
  "mcpServers": {
    "playwright": {
      "url": "http://[::1]:1234/mcp"
    }
  }
}
For legacy SSE transport support, you can use the /sse endpoint instead.
```

pavelfeldman:
I believe it is working as intended, regardless of the bound port we print a config that would work locally. We don't know actual interfaces that are going to be used for the interface bound to 0.0.0.0 anyways.

brianredbeard:
[...]

PR #791, opened 2025-07-29
fix: Don't mask server hostname

brianredbeard:
PR #751 introduced invalid hostnames which are sent to the output, which creates issues with troubleshooting.

This cleans up the incorrect informational messages.

Closes https://github.com/microsoft/playwright-mcp/issues/790


pavelfeldman:
Current behavior is by design. localhost convenience shortcut is orthogonal to the bound interface method.

PR #797, opened 2025-07-30
chore: use pngs by default for screenshots

mxschmitt:
1. Use PNG by default.
1. Increase JPG quality from `50` -> `90`.


src/tools/screenshot.ts
```diff
@@ -24,7 +24,7 @@ import { generateLocator } from './utils.js';
 import type * as playwright from 'playwright';
 
 const screenshotSchema = z.object({
-  raw: z.boolean().optional().describe('Whether to return without compression (in PNG format). Default is false, which returns a JPEG image.'),
+  type: z.enum(['png', 'jpeg']).optional().describe('Image format for the screenshot. Default is png, or inferred from filename extension.'),
```
pavelfeldman: Let the LLM do inference, just say defaults to PNG

src/tools/screenshot.ts
```diff
@@ -52,11 +58,11 @@ const screenshot = defineTabTool({
   },
 
   handle: async (tab, params, response) => {
-    const fileType = params.raw ? 'png' : 'jpeg';
+    const fileType = params.type || 'png';
     const fileName = await outputFile(tab.context.config, params.filename ?? `page-${new Date().toISOString()}.${fileType}`);
     const options: playwright.PageScreenshotOptions = {
       type: fileType,
-      quality: fileType === 'png' ? undefined : 50,
+      quality: fileType === 'png' ? undefined : 80,
```
pavelfeldman: When the size is not a concern (our case), the usual default is 90.

pavelfeldman (approved):

PR #815, opened 2025-08-01
chore: allow to switch between browser connection methods

yury-s:


src/browserServerBackend.ts
```diff
@@ -21,12 +21,19 @@ import { Response } from './response.js';
 import { SessionLog } from './sessionLog.js';
 import { filteredTools } from './tools.js';
 import { packageJSON } from './package.js';
+import { ContextSwitchTool } from './tools/contextSwitch.js';
```
pavelfeldman: I would define this tool inline in this file.

src/browserServerBackend.ts
```diff
@@ -21,12 +21,19 @@ import { Response } from './response.js';
...
 
 import type { BrowserContextFactory } from './browserContextFactory.js';
 import type * as mcpServer from './mcp/server.js';
 import type { ServerBackend } from './mcp/server.js';
 import type { Tool } from './tools/tool.js';
 
+export type FactoryList = {
+  name: string;
```
pavelfeldman: Let's move name and description into the BrowserContextFactory

src/browserServerBackend.ts
```diff
@@ -37,10 +44,16 @@ export class BrowserServerBackend implements ServerBackend {
   private _config: FullConfig;
   private _browserContextFactory: BrowserContextFactory;
 
-  constructor(config: FullConfig, browserContextFactory: BrowserContextFactory) {
+  constructor(config: FullConfig, browserContextFactory: BrowserContextFactory | FactoryList) {
```
pavelfeldman: Let's make it always a list.

src/tools/contextSwitch.ts
```diff
@@ -0,0 +1,61 @@
...
+  capability = 'core' as const;
+
+  schema = {
+    name: 'browser_connect',
+    title: 'Connect to a browser context',
+    description: 'Connect to a browser using one of the available methods',
+    inputSchema: z.object({
+      // method
+    }),
+    type: 'destructive' as const,
```
pavelfeldman: The type of this tool is readOnly as it does not mess with the user data.

src/tools/contextSwitch.ts
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  private _factories: FactoryList;
+  private _onChange: (newFactory: BrowserContextFactory) => void;
+
+  constructor(factories: FactoryList, onChange: (newFactory: BrowserContextFactory) => void) {
+    this._factories = factories;
+    this._onChange = onChange;
+
+    this.schema.description = [
```
pavelfeldman: We would not need it is we moved the switcher tool into the server.

pavelfeldman (approved):

Issue #395, opened 2025-05-10
the snapshot length is too long, add pagination to it


Weixuanf:
the snapshot returned is too long for some website, resulting the LLM token errors, and its a big waste of token.
Please add pagination to all view snapshot methods, so that LLM can retrieve page by page for long website snapshot

pavelfeldman:
We believe that the tokens will get cheaper faster than we can elaborate a meaningful pagination solution.

Weixuanf:
I think it's not only about pricing. The context length has a limit of LLM, if context is too long (and lots of then are irrelevant), it adds lots of noise to LLM and can reduce the quality of its answering I believe. At least for now, when LLM haven't got unlimited context length yet...

ElijahSink:
This explicitly doesn't work at all for Claude Code right now for long pages. I'm getting this response for pretty much all tool usages against a decently long Swagger docs page:

> Error: MCP tool "browser_snapshot" response (71823 tokens) exceeds maximum allowed tokens (25000). Please use pagination, filtering, or limit parameters to reduce the response size.

elyxlz:
this is very annoying, its a simple fix !

pietrodn:
I tried encoding the response in JSON instead of YAML hoping that it would reduce the number of tokens, however the impact was in the order of -0.2% only. :-( 

Removing the `cursor` attribute (`[cursor=pointer]`) that is on nearly every element gives a -12% reduction which is not bad...

cseickel:
I can't get Claude Code to do anything without hitting this error. Seriously, even if the context limit was way higher, 25K+ tokens is crazy high for every interaction. By the time we get to the point that an AI can read 40K tokens on every interaction and not become useless, they will be so different from what we use today that MCPs will be legacy anyhow. It is not at all a matter of expense, it is a matter of being incapable of handling responses this large by the very nature of how they function.

If it returned a link to a temp file, then the AI could grep it intelligently.

NicolasArnouts:
This is actually a huge issue for me, I've been running into this issue a lot.

pjar:
Came here to see if I'm the only one hitting this issue. Playwright MCP is pretty much useless in Claude Code for anything more than reading console.log because of this.

PR #820, opened 2025-08-02
feat: economize on output tokens

pietrodn:
Here are a few optimizations to reduce the number of output token generated:

- Remove the `[cursor=pointer]` attribute that is added on every clickable element. On a news site I tested (`ilpost.it`) this amounted for ~12% of the tokens of the ARIA snapshot.
- When rendering the output of `browser_snapshot`, output the ARIA snapshot before the console messages and other info. By keeping large content at the beginning of the prompt, we are able to exploit [implicit context caching](https://ai.google.dev/gemini-api/docs/caching) on Gemini, potentially saving on costs if the bulk of the snapshot does not change among multiple invocations.
- Omit all the console logs with severity level < `WARNING`, and deduplicate them.

This should at least mitigate the issues of #395.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -16,9 +16,9 @@
 
 import { renderModalStates } from './tab.js';
 
-import type { Tab, TabSnapshot } from './tab.js';
 import type { ImageContent, TextContent } from '@modelcontextprotocol/sdk/types.js';
 import type { Context } from './context.js';
+import type { Tab, TabSnapshot } from './tab.js';
 
 export class Response {
   private _result: string[] = [];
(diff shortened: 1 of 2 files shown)


pietrodn:

src/response.ts
```diff
@@ -143,10 +143,33 @@ ${this._code.join('\n')}
 function renderTabSnapshot(tabSnapshot: TabSnapshot): string {
   const lines: string[] = [];
 
+  // Put first the elements that are less likely to change for the same web page,
+  // to exploit Context Caching whenever it is available.
```
pavelfeldman, line 147: Can you explain how this is helpful? We combine these messages including console entries into a single large history anyways.

src/response.ts
```diff
@@ -143,10 +143,33 @@ ${this._code.join('\n')}
...
+  lines.push(tabSnapshot.ariaSnapshot);
+  lines.push('```');
+
+  const messageSet = new Set<string>();
+
   if (tabSnapshot.consoleMessages.length) {
     lines.push(`### New console messages`);
-    for (const message of tabSnapshot.consoleMessages)
-      lines.push(`- ${trim(message.toString(), 100)}`);
+    for (const message of tabSnapshot.consoleMessages) {
```
pavelfeldman, line 160: LLM might choose to inject console.logs for debugging, would be a shame to not see them in console.

src/tab.ts
```diff
@@ -16,13 +16,21 @@
...
+import { callOnPageNoTrace, waitForCompletion } from './tools/utils.js';
 
 import type { Context } from './context.js';
 
+/**
+ * Recursively traverses a data structure (object or array) and removes the
+ * substring ' [cursor=pointer]' from all string values.
+ */
+function cleanYamlValue(data: any): any {
+  return data.replaceAll(' [cursor=pointer]', '');
```
pavelfeldman, line 31: You are removing the strongest actionability signal from the snapshot, not sure users are going to appreciate the context savings.

PR #836, opened 2025-08-05
chore(extension): show error when connection is rejected due to inact…

yury-s:
…ivity

### DIFF
--- a/extension/src/ui/connect.tsx
+++ b/extension/src/ui/connect.tsx
@@ -122,7 +122,7 @@ const ConnectApp: React.FC = () => {
   }, []);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


pavelfeldman (approved):

extension/src/ui/connect.tsx
```diff
@@ -121,6 +121,13 @@ const ConnectApp: React.FC = () => {
     setStatus({ type: 'error', message: 'Connection rejected. This tab can be closed.' });
   }, []);
 
+  useEffect(() => {
```
dgozman, line 124: Note: you are supposed to return a function that removes the event listener.

PR #838, opened 2025-08-06
feat: Add configurable default timeout

zhurong666:


zhurong666:
> @microsoft-github-policy-service agree

PR #847, opened 2025-08-07
chore(extension): explicitly detach from debugger when connection closes

yury-s:


extension/src/relayConnection.ts
```diff
@@ -79,6 +79,7 @@ export class RelayConnection {
     this._closed = true;
     chrome.debugger.onEvent.removeListener(this._eventListener);
     chrome.debugger.onDetach.removeListener(this._detachListener);
+    chrome.debugger.detach(this._debuggee).then(() => {});
```
dgozman:
```suggestion
    chrome.debugger.detach(this._debuggee).catch(() => {});
```

dgozman (approved):

PR #854, opened 2025-08-08
chore: mcp backend switcher

yury-s:


pavelfeldman:
I would do something like this

Screenshot 2025-08-08 at 5 54 15 PM

where each client is a connected server. that way we multiplex on the mcp level and the facade server can switch between the tools, partially or altogether.

src/inProcessClient.ts
```diff
@@ -0,0 +1,54 @@
...
+    this._contextFactory = contextFactory;
+    this._config = config;
+  }
+
+  async create(): Promise<Client> {
+    const client = new Client({
+      name: this.name,
+      version: packageJSON.version
+    });
+    client.setRequestHandler(PingRequestSchema, async () => ({}));
```
pavelfeldman: We are passing runHeartbeat === false, why supporting ping in-process?

src/program.ts
```diff
@@ -78,11 +83,16 @@ program
         return;
       }
 
-      const browserContextFactory = contextFactory(config);
-      const factories: FactoryList = [browserContextFactory];
-      if (options.connectTool)
-        factories.push(createExtensionContextFactory(config));
-      const serverBackendFactory = () => new BrowserServerBackend(config, factories);
+      let serverBackendFactory: () => ServerBackend;
```
pavelfeldman: we have ServerBackendFactory

src/mcp/proxyBackend.ts
```diff
@@ -0,0 +1,137 @@
...
+  private _clientFactories: ClientFactoryList;
+  private _currentFactory: ClientFactory;
+  private _currentClient: Client | undefined;
+  private _contextSwitchTool: Tool<any>;
+  private _server: Server | undefined;
+  private _tools: ToolSchema<any>[] = [];
+
+  constructor(clientFactories: ClientFactoryList) {
+    this._clientFactories = clientFactories;
+    this._currentFactory = this._clientFactories[0];
```
pavelfeldman: _currentFactory is unused

PR #864, opened 2025-08-11
chore: introduce check-deps

pavelfeldman:


utils/check-deps.js
```diff
@@ -0,0 +1,179 @@
...
+import path from 'path';
+
+const __dirname = path.dirname(new URL(import.meta.url).pathname);
+
+const depsCache = {};
+const packageRoot = path.resolve(__dirname, '..');
+
+async function checkDeps() {
+  const deps = new Set();
+  const src = path.join(packageRoot, 'src');
```
yury-s, line 32: check  extension/src too ?

pavelfeldman: I consider it a separate package, we can run check-deps in it too when you have multiple folders

PR #868, opened 2025-08-12
chore(vscode): add vscode client factory

Skn0tt:


src/vscode/host.ts
```diff
@@ -0,0 +1,52 @@
...
+  name = 'vscode';
+  description = 'Connect to a browser running in the Playwright VS Code extension';
+
+  constructor(private readonly _config: FullConfig) {}
+
+  async create(server: Server, options: any): Promise<Client> {
+    if (typeof options.connectionString !== 'string')
+      throw new Error('Missing options.connectionString');
+    if (typeof options.lib !== 'string')
+      throw new Error('Missing options.library');
```
yury-s:
```suggestion
      throw new Error('Missing options.lib');
```

src/vscode/host.ts
```diff
@@ -0,0 +1,52 @@
...
+      command: process.execPath,
+      cwd: process.cwd(),
+      args: [
+        new URL('./main.js', import.meta.url).pathname,
+        JSON.stringify(this._config),
+        options.connectionString,
+        options.lib,
+      ],
+    }));
+    await client.ping();
```
yury-s: we shouldn't need it with stdio transport, I was going to remove it in inProcessClient as well.

src/vscode/main.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const playwright = await import(lib).then(mod => mod.default ?? mod) as typeof import('playwright');
+
+class VSCodeBrowserContextFactory implements BrowserContextFactory {
+  name = 'vscode';
+  description = 'Connect to a browser running in the Playwright VS Code extension';
+
+  async createContext(clientInfo: ClientInfo, abortSignal: AbortSignal): Promise<{ browserContext: BrowserContext; close: () => Promise<void>; }> {
+    connectionString.searchParams.set('launch-options', JSON.stringify({
+      ...config.browser.launchOptions,
```
yury-s: nit: why mix them up rather than keep in separate fields?

src/vscode/main.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { BrowserContext } from 'playwright-core';
+import { FullConfig } from '../config.js';
+import * as mcpServer from '../mcp/server.js';
+import { BrowserServerBackend } from '../browserServerBackend.js';
+import { BrowserContextFactory, ClientInfo } from '../browserContextFactory.js';
+
+const config: FullConfig = JSON.parse(process.argv[2]);
```
yury-s: nit: consider wrapping these logic into main() function and passing the arguments to the factory explicitly.

src/vscode/main.ts
```diff
@@ -0,0 +1,58 @@
...
+  description = 'Connect to a browser running in the Playwright VS Code extension';
+
+  async createContext(clientInfo: ClientInfo, abortSignal: AbortSignal): Promise<{ browserContext: BrowserContext; close: () => Promise<void>; }> {
+    connectionString.searchParams.set('launch-options', JSON.stringify({
+      ...config.browser.launchOptions,
+      ...config.browser.contextOptions,
+      userDataDir: config.browser.userDataDir,
+    }));
+
+    const browser = await playwright.chromium.connect(connectionString.toString());
```
yury-s: Is it supposed to work with WebKit and Firefox?

yury-s (commented):
Add a basic test for the new mode?

src/vscode/main.ts
Skn0tt: yes, I made that clearer in the code

src/vscode/main.ts
Skn0tt: because that's how `launchPersistentContext` works - I updated the code to make that clearer

pavelfeldman (approved):

PR #884, opened 2025-08-13
chore: bind factories later

Skn0tt:
Allows setting factories based on whether the client is VS Code.


pavelfeldman:
Looks stale, also, can we just pass array from the program?

Skn0tt:
> Looks stale

I think it's better if I hold off for a couple days until this all has settled a bit :)

> can we just pass array from the program?

Don't see how - in `program.ts`, we don't know the client name yet. I want to be able to define the provider list based on the client name.

PR #897, opened 2025-08-15
docs(extension): update build and install instructions

schealex:
Current readme instructions are not working. They are missing the build step (you need to add the `dist` folder as extension to chome).

I've added Node as requirement, optional clone instead of download and build instructions


schealex:

pavelfeldman:
We don't really want users to build the extension.

PR #916, opened 2025-08-19
Update README from mcpServers to servers

GeekTrainer:
Because the section in VS Code is titled `servers`, this allows for a copy/paste into an empty `mcp.json` file

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ First, install the Playwright MCP server with your client.
 
 ```js
 {
-  "mcpServers": {
+  "servers": {
     "playwright": {
       "command": "npx",
       "args": [


README.md
```diff
@@ -25,7 +25,7 @@ First, install the Playwright MCP server with your client.
 
 ```js
 {
-  "mcpServers": {
+  "servers": {
```
pavelfeldman, line 28: This is a generic config and the industry has converged on `mcpServers`.

GeekTrainer: Thanks for the context! I'd still say, given the nascent nature of MCP and that VS Code still uses `servers`, it'd be worth a note at the very least to help new users.

PR #932, opened 2025-08-22
Add sentence to describe latest changes

JoannaaKL:
Recent [change](https://github.com/microsoft/playwright-mcp/pull/849) changed the behaviour of `browser_take_screenshot`. This pr adds tool description that should help agent pick correct tool parameters.


pavelfeldman:
Could you share a context where this helps? We are being cautious wrt increasing the tool header sizes.

JoannaaKL:
Hey @pavelfeldman 👋  Mcp tools are used primarily by AI agents. They make the decision which tool to execute based on tool parameters and their descriptions. Recent changes break the expected tool behaviour and need to be documented in the `browser_take_screenshot` description. 

To give you more context which flows that are currently broken:
- agent executes `browser_take_screenshot`
- if the image size exceeds max dimensions, it will be automatically resized
With the recent change tool call won't return any image but there won't be any error returned as well. The agent can't really know why it failed, so it has no clue it could retry without the `fullPage` parameter.

The exact wording can be different but I think it is important to note that including `fullPage` will not return the image data from the tool call.

pavelfeldman:
I understand, I was asking for exact context of this misconception on LLM's side, could you share a trajectory?

pavelfeldman:
It is best to start with filing an issue!

PR #951, opened 2025-08-26
chore: introduce verification tools

pavelfeldman:


src/tools/verify.ts
```diff
@@ -0,0 +1,149 @@
...
+
+const verifyElement = defineTabTool({
+  capability: 'verify',
+  schema: {
+    name: 'browser_verify_element_visible',
+    title: 'Verify element visible',
+    description: 'Verify element is visible on the page',
+    inputSchema: z.object({
+      role: z.string().describe('ROLE of the element. Can be found in the snapshot like this: \`- {ROLE} "Accessible Name":\`'),
+      accessibleName: z.string().describe('ACCESSIBLE_NAME of the element. Can be found in the snapshot like this: \`- role "{ACCESSIBLE_NAME}"\`'),
```
yury-s, line 31: make it optional?

src/tools/verify.ts
```diff
@@ -0,0 +1,149 @@
...
+  handle: async (tab, params, response) => {
+    const locator = await tab.refLocator({ ref: params.ref, element: params.element });
+    const locatorSource = `page.${await generateLocator(locator)}`;
+    if (params.type === 'textbox' || params.type === 'slider' || params.type === 'combobox') {
+      const value = await locator.inputValue();
+      if (value !== params.value) {
+        response.addError(`Expected value "${params.value}", but got "${value}"`);
+        return;
+      }
+      response.addCode(`await expect(${locatorSource}).toHaveValue(${javascript.quote(params.value)});`);
```
yury-s, line 130: This doesn't match what the tool does, let's update the tool to be auto-waiting too?

yury-s (approved):

src/tools/verify.ts
pavelfeldman: It is not practical to refer to an element by role, there is name in the snapshot already.

src/tools/verify.ts
pavelfeldman: It can't auto-wait since it'll take the value from the existing snapshot.

PR #964, opened 2025-08-28
Fix `browser_tabs` select action with index 0 failing due to falsy check

healtheloper:
The `browser_tabs` fails when trying to select tab with index 0 due to an incorrect falsy
  value check in the code. The current implementation likely uses `if(index)` which treats `index: 0` as
  false, causing the "Tab index is required" error even when index 0 is explicitly provided.


healtheloper:
Added a test. Please let me know if anything needs improvement. Thank you.

healtheloper:

pavelfeldman (approved):

PR #970, opened 2025-08-29
fix: throw if an instance of the browser is already running

yury-s:


src/browserContextFactory.ts
```diff
@@ -189,7 +190,13 @@ class PersistentContextFactory implements BrowserContextFactory {
       } catch (error: any) {
         if (error.message.includes('Executable doesn\'t exist'))
           throw new Error(`Browser specified in your config is not installed. Either install it (likely) or change the config.`);
+        if (error.message.includes('A copy of Firefox is already open. Only one copy of Firefox can be open at a time.'))
+          throw new Error(`Browser is already running for '${userDataDir}', use --isolated to run multiple instances of the same browser or wait for the existing one to close.`);
         if (error.message.includes('ProcessSingleton') || error.message.includes('Invalid URL')) {
+          const execPath = this.config.browser.launchOptions.executablePath ?? getBrowserExecPath(this.config.browser.launchOptions.channel ?? browserType.name());
```
pavelfeldman: That would be too late. You should not start the browser if you know it is running, and you should still wait for 5 seconds while checking.

pavelfeldman:
Sorry you are experiencing moving ground here. We upstreamed the code to the playwright monorepo and will soon have instructions for contributing in the new model ready.

PR #973, opened 2025-08-29
chore: use mcp sdk via playwright

pavelfeldman:


package.json
```diff
@@ -53,6 +52,7 @@
     "@anthropic-ai/sdk": "^0.57.0",
     "@eslint/eslintrc": "^3.2.0",
     "@eslint/js": "^9.19.0",
+    "@modelcontextprotocol/sdk": "^1.16.0",
```
dgozman, line 53: Shouldn't `zod` and `zod-to-json-schema` also move to `devDependencies`, since these come from the same bundle as well?

Issue #983, opened 2025-09-02
[CDP Connection] Support storage‑state configuration when connecting to a browser via a CDP endpoint


Ivan97:
## Description
We would like the Playwright MCP to support loading a storage‑state file when it connects to a browser through a CDP endpoint.
At the moment the `--cdp-endpoint` flag works, but there is no way to tell the remote context to initialise from a previously saved storageState (cookies, localStorage, etc.). Adding this capability would make it possible to reuse an authenticated session without having to keep a persistent Chrome profile.

## Desired behaviour

When the MCP transport starts a Playwright process with a CDP connection, it should accept a --storage-state argument (or an equivalent API option) and apply that state to the newly created browser context.

Illustrative Kotlin configuration using LangChain4j (the pattern we would like to be supported):

```kotlin
val transport = StdioMcpTransport.Builder()
    .command(
        listOf(
            "npx",
            "@playwright/mcp@latest",
            "--isolated",                                   // start a fresh, isolated context
            "--storage-state=/home/ubuntu/workspace/state.json", // <-- new flag
            "--cdp-endpoint=http://xx.xx.xx.xx:9222"        // connect to remote Chrome
        )
    )
    .logEvents(true)
    .build()
```

With the flag above we expect the following sequence internally:

1. `playwright.chromium.connectOverCDP('http://xx.xx.xx.xx:9222')` is invoked.
2. A new browser context is created with storageState loaded from the JSON file located at /home/ubuntu/workspace/state.json.
3. All subsequent pages opened inside that context inherit the cookies/localStorage stored in the file, so the session is already logged‑in.

## Expected CLI signature

```
npx @playwright/mcp@latest \
    --isolated \
    --cdp-endpoint=<url> \
    --storage-state=<path-to-json>
```

- `--storage-state` should accept a path to a JSON file produced by `context.storageState({ path: ... })`.
- If the file does not exist or is malformed, the CLI should emit a clear error and abort the launch.
- The flag should work in all language bindings (JS/TS, Python, .NET, Java) that rely on the MCP transport.

pavelfeldman:
This falls outside of the scope for Playwright

Ivan97:
@pavelfeldman Hi, may I ask why?

PR #984, opened 2025-09-02
Adding support for storage‑state configuration when connecting to a browser via a CDP endpoint

Ivan97:
related to [playwright-mcp/issues/983](https://github.com/microsoft/playwright-mcp/issues/983)


Ivan97:

pavelfeldman:
This falls outside of the scope for Playwright

PR #991, opened 2025-09-03
feat: add browser_save_blob tool and improve file saving descriptions

igormf:
## Summary
  - Adds new `browser_save_blob` tool to handle the classic Playwright problem of saving blob URLs (e.g., PDFs opened in browser viewers)
  - Improves file saving descriptions across all tools for better LLM understanding

  ## Changes
  ### New Tool: `browser_save_blob`
  - Automatically detects blob URLs in current page, iframes, and embed elements
  - Supports manual blob URL specification with filename parameter
  - Determines file extension from content type (PDF, PNG, JPG, etc.)
  - Uses same secure file saving mechanism as other tools

  ### Improved File Saving Descriptions
  - Updated `browser_pdf_save`, `browser_take_screenshot`, and `browser_save_blob` descriptions
  - Clarifies that files are saved to MCP server's configured output directory only
  - Explains that arbitrary paths cannot be specified for security
  - All tools now return full paths in results for LLM assurance

  **Rationale**: Testing revealed that most LLMs (Claude, Gemini, Codex) got confused about where screenshots and PDF saves were being sent. They attempted to use        
  full path parameters, which just ended up becoming filenames with dashes instead of slashes, causing confusion about file locations.

  ## Test Coverage
  - ✅ Manual blob URL with filename - Creates text blob, saves with specified filename, verifies content
  - ✅ Auto-detection with iframe - Creates blob in iframe, auto-detects and saves
  - ✅ Error handling for missing blob URLs
  - ✅ Error handling for invalid non-blob URLs
  Tests follow same patterns as existing PDF/screenshot tests

  ## Test plan
  - [x] Tool compiles and builds successfully
  - [x] Tool is properly registered in the tools system
  - [x] Descriptions are clear and consistent across all file-saving tools
  - [x] Comprehensive test coverage with actual file verification

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import blob from './tools/blob';
 import common from './tools/common';
 import console from './tools/console';
 import dialogs from './tools/dialogs';
@@ -36,6 +37,7 @@ import type { Tool } from './tools/tool';
 import type { FullConfig } from './config';
 
 export const allTools: Tool<any>[] = [
+  ...blob,
   ...common,
   ...console,
   ...dialogs,
--- a/src/tools/blob.ts
+++ b/src/tools/blob.ts
@@ -1,0 +1,156 @@
+/**
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


igormf:

pavelfeldman:
Thanks for the PR! It is best to start with filing an issue and seeing if there is positive feedback from the maintainers!

igormf:
this would fix #1006 and #1002

Issue #989, opened 2025-09-03
Feature Request: CDP Authentication Headers Support


phatpham-katalon:
## Prolem : 
Currently, MCP Playwright supports connecting to existing browser instances via CDP endpoints using the `CdpContextFactory`, but it does not support custom authentication headers required by some browser automation services like AgentCore Browser. 

## Feature Request:
Add support for custom HTTP headers in CDP connections to enable authentication with services that require header-based authentication.  

## More Detail: 

Extend the `CdpContextFactory` to support additional connection parameters:  
  
1. **Configuration Schema Extension**: Add support for CDP headers in the config  
   ```json  
   {  
     "cdpEndpoint": "ws://localhost:9222",  
     "cdpHeaders": {  
       "Authorization": "Bearer token",  
       "X-Custom-Auth": "value"  
     }  
   }

pavelfeldman:
This has landed and will be a part of 0.0.37

PR #1011, opened 2025-09-07
 feat: CDP Authentication Headers Support for issue #989

phatpham-katalon:
- Added -cdp-headers as a parameter when connect to an external browser (such as Bedrock Agent Core or other browser need headers for authentication) 
- Added test for this


phatpham-katalon:

PR #1037, opened 2025-09-14
docs: fix cursor mcp link redirect error

zeyios:
The current link redirected to the wrong page, likely because the `cursor://xxxx` scheme is not recognized by Markdown. I updated the link to use `https://xxxx` instead.

Previous: 
[Install in Cursor](cursor://anysphere.cursor-deeplink/mcp/install?name=Playwright&config=eyJjb21tYW5kIjoibnB4IEBwbGF5d3JpZ2h0L21jcEBsYXRlc3QifQ%3D%3D)

Fixed:
[Install in Cursor](https://cursor.com/en/install-mcp?name=Playwright&config=eyJjb21tYW5kIjoibnB4IEBwbGF5d3JpZ2h0L21jcEBsYXRlc3QifQ%3D%3D)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ For more information, see the [Codex MCP documentation](https://github.com/opena
 
 #### Click the button to install:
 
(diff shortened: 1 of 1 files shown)


msfeldstein:
Hi i also made this change in https://github.com/microsoft/playwright-mcp/pull/1041
as well as adding the `-y` flag which is necessary for new users

msfeldstein:
Thanks for finding it!

PR #1041, opened 2025-09-15
docs: Fix cursor install instructions

msfeldstein:
The MCP server won't run properly if you haven't run the npx before, so you need to add `-y` to run it
Also update the deeplink to use -y and use the https redirect since you can't deeplink cursor:// on github

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ For more information, see the [Codex MCP documentation](https://github.com/opena
 
 #### Or install manually:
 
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Someone beat you to it: https://github.com/microsoft/playwright-mcp/pull/1037. I don't think you need -y as Cursor should be running it in a non-interactive environment where -y is default.

msfeldstein:
mm you do need -y, i can see about making it not necessary, but we see the same behavior in claude-code as well

msfeldstein:
ah ok i see i can pass NPM_CONFIG_YES=true to make that not happen and i can get this out in the next build of cursor

PR #1049, opened 2025-09-17
chore: mark v0.0.38

pavelfeldman:


extension/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "Playwright MCP Bridge",
-  "version": "0.0.37",
```
yury-s, line 4: Don't bump extension?

pavelfeldman: I will as long as it matches!

PR #1137, opened 2025-10-15
Fix/streamable http session management

BrentGTR:
Updated the dockerfile to allow for streamable-http transport. 

When combined with the playwright-source PR, these two changes will fix the "navigate to google, type banana into the search bar, click search and take a screen shot of the result" scenario without a session being dropped when using it in a pod.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "Playwright MCP Bridge",
-  "version": "0.0.41",
+  "version": "0.0.42",
   "description": "Share browser tabs with Playwright MCP server",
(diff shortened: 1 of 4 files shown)


pavelfeldman:
I don't think users expect this breaking change to happen. Please start with filing a feature request or a bug.

BrentGTR:
@pavelfeldman this fix is to resolve an issue that I am experiencing with the mcp deployed to a pod. The agent gets the initial "navigate" request through but the session is deleted and consecutive actions fail. This works locally, through stdio, but fails due to the session closing before subsequent actions can be executed. 

I'll file a bug report with detailed reproduction steps. The issue is:
- **Environment**: MCP deployed in Kubernetes pod
- **Symptom**: First request succeeds, subsequent requests fail with "Session not found"
- **Root Cause**: HTTP transport immediately deletes sessions on disconnect
- **Impact**: Multi-step browser automation fails in containerized environments

The fix (delayed session cleanup) resolves this but changes behavior, so it needs proper review and discussion.

I'll create the bug report with:
1. Exact reproduction steps
2. Environment details (pod vs local)
3. Expected vs actual behavior
4. Proposed solution with impact analysis

Thanks for the guidance on proper process!

BrentGTR:
Bug created - https://github.com/microsoft/playwright-mcp/issues/1140. Thanks, @pavelfeldman!

PR #1197, opened 2025-11-14
Add step for publishing to MCP Registry

joelverhagen:
This adds a new step for the `release` event to publish to the Official MCP Registry.

Summary of changes:
- Add `server.json` which is the MCP Server manifest
  - Used metadata from the GH MCP Registry
  - Used the name `com.microsoft/playwright-mcp`
- Add `publish-release-mcp-registry` job that runs manually
   - Install Microsoft Go for compliant Go crypto
   - Log in to Azure for Key Vault access
   - Publish to the MCP Registry in the `microsoft.com` namespace.
- Add `mcpName` to the `package.json` which the MCP Registry validates.

The publishing requires some secret variables:
- `AZURE_MCP_REGISTRY_CLIENT_ID` - service principal client ID (app ID) with Get Key and Sign access to a Key Vault
- `AZURE_MCP_REGISTRY_TENANT_ID` - tenant ID for the service principal (should be a production tenant)
- `AZURE_MCP_REGISTRY_SUBSCRIPTION_ID` - Azure subscription ID holding the Key Vault
- `KV_NAME` - the Key Vault name, e.g. `mykeyvault`
- `KV_KEY_NAME` - the name of a Key in the Key Vault, needs to be ECDSA P-384.

I can provide internal documentation for how to set up the service principal (managed identity is easiest), vault, and key as well as registering the key so it can be used for the `microsoft.com` namespace.

**Note:** after the new job is proven to work as expected, we can switch it to automatically running after the `publish-release-npm` job on the `release` event.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -149,8 +149,7 @@ jobs:
             gh release upload ${{github.event.release.tag_name}} ./extension/playwright-mcp-extension-${{ steps.get-version.outputs.version }}.zip
 
   publish-release-mcp-registry:
-    if: github.event_name == 'release'
-    needs: publish-release-npm
+    if: github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     environment: allow-mcp-registry-publishing
     permissions:


.github/workflows/publish.yml
```diff
@@ -147,3 +147,60 @@ jobs:
...
+      id-token: write # Needed for GitHub OIDC authentication
+    steps:
+      - uses: actions/checkout@v5
+      - name: Clone MCP Registry and build publisher tool
+        shell: pwsh
+        run: |
+          cd ${{ runner.temp }}
+
+          # Install Microsoft Go
+          & ${{ github.workspace }}/.mcp/go-install.ps1
```
Skn0tt: Can you elaborate on why we need a special Microsoft build? My understanding is that the repo we're building isn't Microsoft-specific, so why can't we use standard Go?

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@playwright/mcp",
   "version": "0.0.46",
   "description": "Playwright Tools for MCP",
+  "mcpName": "com.microsoft/playwright",
```
Skn0tt: we probably want this called `playwright-mcp`. we already have a different MCP server distributed as part of our test runner, and we don't want to namesquat ourselves for future potential products.
```suggestion
  "mcpName": "com.microsoft/playwright-mcp",
```

.mcp/server.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://static.modelcontextprotocol.io/schemas/2025-10-17/server.schema.json",
+  "name": "com.microsoft/playwright",
```
Skn0tt: see below
```suggestion
  "name": "com.microsoft/playwright-mcp",
```

.mcp/server.json
```diff
@@ -0,0 +1,21 @@
...
+  "title": "Playwright",
+  "description": "Automate web browsers using accessibility trees for testing and data extraction.",
```
Skn0tt, line 5: just to double-check, we can change this freely in future versions, right? Does the CLI source our README from somewhere, or where does further documentation come from?

Skn0tt (commented):
Thanks! Left a couple of comments from a first cursory look.

package.json
joelverhagen: The name here is for the MCP Registry, so I think the `-mcp` suffix would be redundant. If this is not the "main" Playwright MCP server, then I think a different qualifier would make sense. This is like suffixing an npm package name with `npm` or `package`. But it's up to you.

.mcp/server.json
joelverhagen: My understanding is that GitHub MCP Registry uses the README from the repository as it is consuming from the Official MCP Registry.

Further documentation should be specified on the `websiteUrl` field, which I omitted due to an aggressive validation on the MCP Registry, which I am submitting a PR to fix (see https://github.com/modelcontextprotocol/registry/issues/768).

.github/workflows/publish.yml
```diff
@@ -147,3 +147,60 @@ jobs:
...
+          $serverJson = Get-Content ./.mcp/server.json | ConvertFrom-Json
+          $serverJson.version = $version
+          foreach ($package in $serverJson.packages) {
+            $package.version = $version
+          }
+          $serverJson | ConvertTo-Json -Depth 10 | Set-Content ./.mcp/server.json
+      - name: Azure Login via OIDC
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.KV_CLIENT_ID }}
```
Skn0tt: let's reuse the same identity we have for the docker image connection above

.github/workflows/publish.yml
joelverhagen: When reviewing this flow with compliance experts internally, we agreed that it would be more prudent to build the publisher tool with Microsoft Go rather than using the binaries from GitHub Release.

.mcp/server.json
Skn0tt: Pulling README sounds good for now. If your `websiteUrl` change ships, let's point to `https://github.com/microsoft/playwright-mcp` in `websiteUrl` here. Maybe we want to change where it points in the future.

package.json
Skn0tt: We don't yet have a notion of a "main" server, and we don't know if we'll ever have. Currently, `Playwright MCP` is the product name we have. So `playwright-mcp` would be good here I think, let me double check with the rest of the team.

.mcp/server.json
joelverhagen: Oh and yes, you can change it freely in the future. The "identity" of this listing is the `name`. As long as you bump the `version`, you can change any of the other metadata.

package.json
joelverhagen: Note that the listing is currently `Playwright` not `Playwright MCP` on the GH MCP Registry right now:
image
We can change it for sure, I just wanted to provide the baseline that I was aware of.

joelverhagen: (to be clear the `title` will be used for display here, not the `name`, but I noticed the `names` are generally not peppered with `mcp` suffixes -- just providing you more context to make the decision you think is best)

.github/workflows/publish.yml
Skn0tt: I don't like the idea of having a stray `go-install.ps1` file in our repo that we have to keep up-to-date. Is there maybe a microsoft-maintained Github action we can use to install Microsoft Go?

Skn0tt:
I setup a Keyvault and key, and populated the `KV_KEY_NAME` and `KV_NAME` secrets.

package.json
Skn0tt: We just discussed this in our daily, let's go with `playwright-mcp`.

.github/workflows/publish.yml
yury-s: I agree, having another 1k lines of the scripts that we'll have to maintain just for publishing doesn't sound great. Can the tool be turned into a github action, so that each client like us doesn't have to build it?

.github/workflows/publish.yml
yury-s: As discussed in the meeting today, let's not mix unrelated entries. I've created a separate app registration for the mcp registry publishing to make things clear.

yury-s: I've added 
AZURE_MCP_REGISTRY_CLIENT_ID,
AZURE_MCP_REGISTRY_SUBSCRIPTION_ID, 
AZURE_MCP_REGISTRY_TENANT_ID
secrets that you can use for publishing. The app registration has access to the KeyVault that Simon created earlier.

.github/workflows/publish.yml
```diff
@@ -147,3 +147,60 @@ jobs:
            GITHUB_TOKEN: ${{ github.token }}
         run: |
             gh release upload ${{github.event.release.tag_name}} ./extension/playwright-mcp-extension-${{ steps.get-version.outputs.version }}.zip
+
+  publish-release-mcp-registry:
+    if: github.event_name == 'release'
+    needs: publish-release-npm
+    runs-on: ubuntu-latest
```
yury-s, line 153: We'll need to add the following environment to match one in the app registration:

```
  environment: allow-mcp-registry-publishing
```

.github/workflows/publish.yml
joelverhagen: There is no GitHub Action for Microsoft Go, yet (https://github.com/microsoft/go/issues/484).

I have switched from a checked in `go-install.ps1` to using a fixed version of their `goinstallscript` helper tool. This could be `@latest` version also if you want to float up to latest acquisition script (I don't recommend that). Until they have a fixed CDN URL you can download the script from (https://github.com/microsoft/go-lab/issues/273) or until the centralized pipeline is ready with our internal release service, I think this is a reasonable approach. FWIW, they did recommend checking in the `go-install.ps1` script but if you want to avoid that, this approach will still provide stable Microsoft Go acquisition script.

We could also have a `go.mod` with `goinstallscript` declared and get dependabot on that, but I think it's probably not worth it since this is a short term solution.

package.json
joelverhagen: Done.

.mcp/server.json
joelverhagen: My PR to MCP Registry is approved by one person but not merged or deployed. I recommend we proceed as-is and come back later with tweaks.

.github/workflows/publish.yml
joelverhagen: Updated.

.github/workflows/publish.yml
joelverhagen: Done.

.github/workflows/publish.yml
```diff
@@ -147,3 +147,62 @@ jobs:
...
+    environment: allow-mcp-registry-publishing
+    permissions:
+      contents: read
+      id-token: write # Needed for GitHub OIDC authentication
+    steps:
+      - uses: actions/checkout@v5
+      - name: Clone MCP Registry and build publisher tool
+        shell: pwsh
+        run: |
+          cd ${{ runner.temp }}
```
yury-s, line 163: Can we just checkout and build in the default workspace?

.github/workflows/publish.yml
```diff
@@ -147,3 +147,62 @@ jobs:
...
+          git clone --branch "v1.3.7" https://github.com/modelcontextprotocol/registry
+          cd registry
+          go build -o ${{ runner.temp }}/mcp-publisher ./cmd/publisher
+
+          # show help for the tool to ensure it's working
+          ${{ runner.temp }}/mcp-publisher --help
+      - name: Update server.json version
+        shell: pwsh
+        run: |
+          $packageJson = Get-Content ./package.json | ConvertFrom-Json
```
yury-s: I'd have it updated manually when we bump package.json version to keep things simple. We anyway to do search-and-replace of old->new version in the project and this version should always match package.json.

.mcp/server.json
joelverhagen: The website URL issue is fixed.

.github/workflows/publish.yml
joelverhagen: I thought it was best to keep the tool compilation out of the main workspace since it's unrelated to the MCP package content. I feel like there is a small risk of the npm packaging at some point including a `registry` subdirectory if the build changes. But if you feel strongly, we can do that.

.github/workflows/publish.yml
joelverhagen: Sounds good.

.github/workflows/publish.yml
yury-s: We currently have `**/*` in .npmignore, so basically there is explicit list of included files. I just want to keep the script as simple as possible, but I see your point about packaging, so we can keep it that way too, no strong opinion.

joelverhagen:
@yury-s, thanks for the approval. Any input on this point?

> **Note:** if we need to iterate on the MCP Registry publish step, it might be worth allowing the `publish-release-mcp-registry` step to execute without the `publish-release-npm` and allow it to run on `workflow_dispatch`. Alternatively, we could allow the `npm publish` step to fail with "duplicate version", making the npm publish step idempotent.

joelverhagen:
Currently the public key is not declared on `microsoft.com`. I sent an email to @Skn0tt on how to do this, so the record is owned by your team.

yury-s:
> Currently the public key is not declared on `microsoft.com`. I sent an email to @Skn0tt on how to do this, so the record is owned by your team.

Can you cc me as well, I may be able to do it earlier?


> @yury-s, thanks for the approval. Any input on this point?
> 
> > **Note:** if we need to iterate on the MCP Registry publish step, it might be worth allowing the `publish-release-mcp-registry` step to execute without the `publish-release-npm` and allow it to run on `workflow_dispatch`. Alternatively, we could allow the `npm publish` step to fail with "duplicate version", making the npm publish step idempotent.

I guess you wan to give it try without waiting for the next mcp release. We can allow it to run on `workflow_dispatch` for now and once you tested it, we can restore the dependency on the npm publish step.

PR #1202, opened 2025-11-17
feat: add --always-allow flag to bypass extension approval dialog

aj47:
I use the extension when im away from computer. there has been recent changes that require the user to approve every session. this PR is to enable a way to always approve all extension sessions

This commit adds a new --always-allow command-line flag that allows users to bypass the browser extension connection approval dialog on each session.

Changes:
- Added CLI wrapper in cli.js to intercept --always-allow flag
- Modified extension to recognize 'always-allow' as a special token value
- Updated documentation in README.md and extension/README.md
- Updated extension UI to show the --always-allow option
- Added test case for the new flag

The flag sets PLAYWRIGHT_MCP_EXTENSION_TOKEN=always-allow internally, providing a cleaner API than requiring users to set environment variables.

Usage:
  npx @playwright/mcp@latest --extension --always-allow

Security note: This flag should only be used in trusted development environments as it allows any MCP client to connect without approval.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,21 @@ state [here](https://playwright.dev/docs/auth).
 
 The Playwright MCP Chrome Extension allows you to connect to existing browser tabs and leverage your logged-in sessions and browser state. See [extension/README.md](extension/README.md) for installation and setup instructions.
 
+To bypass the connection approval dialog on each session, you can use the `--always-allow` flag:
+
+```json
+{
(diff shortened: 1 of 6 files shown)


pavelfeldman:
Please file an issue instead

aj47:
thank you, i have made https://github.com/microsoft/playwright-mcp/issues/1221

PR #1283, opened 2025-12-27
feat(extension): add custom token option for automation

### LINKED ISSUE #1181: [Feature Request] - Constant Token Option for Chrome Extension
The chrome plugin generates a token, which means the browser must be up, I have to fetch the token. I'm trying to fully automate certain pipelines. It would be useful.

AdamParker19:
Adds the ability to set a constant/custom authentication token instead of using the randomly generated one. This enables fully automated pipelines where the token can be set ahead of time.

- Add "Set custom token" expandable section to status page
- Validate token length (min 16 chars) and allowed characters
- Persist custom token to localStorage

Fixes #1181

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,3 +1,3 @@
 # Where is the source?
 
-Playwright MCP source code is located in the Playwright monorepo. Please refer to the contributor's guide in [CONTRIBUTING.md](../CONTRIBUTING.md) for more details.
+Playwright MCP source code is located in the [Playwright monorepo](https://github.com/microsoft/playwright/blob/main/packages/playwright/src/mcp). Please refer to the contributor's guide in [CONTRIBUTING.md](../CONTRIBUTING.md) for more details.


yury-s:
This is just marginally better compared to when the token is generated by the extension. You still have to start the browser and set the token somehow before you can connect to the extension. If you set the token via an existing Playwright connection, why not just copy generated token from the extension using the same Playwright session and then pass the token to the MCP server (yes, it will need to be launched after the browser has been launched once)? Perhaps I am missing something, but how can you set the custom token automatically without launching the browser?

AdamParker19:
Thanks for the feedback @yury-s 
You're right to point out that this still requires launching the browser once to set the token but the looking at the bright side we get the following :-

1. One-time setup: User launches the browser once, sets their custom token in the extension UI
2. Persistent token: The token is stored in `localStorage` and persists across browser restarts
3. Reuse:  Subsequent MCP sessions can use the same token without needing to interact with the extension again
This is marginally better for scenarios where users want predictable, consistent tokens across their development environment rather than random ones that change.
---
However, I agree this doesn't fully solve the "zero browser interaction" automation case. 
### Proposed Enhancement
Would it be acceptable to add support for setting the token via:
1. Config file - Extension reads token from a file like `~/.playwright-mcp/extension-token` on startup
2. Chrome extension sync storage - Using `chrome.storage.local` which persists across reinstalls
3. Environment variable - Read from `PLAYWRIGHT_MCP_EXTENSION_TOKEN` when the extension initializes (if possible via extension APIs)

Which approach would align best with the project's design philosophy? I'm happy to update the PR based on your guidance I am really interested in contributing to this project :)

yury-s:
I’m not convinced this meaningfully improves the situation.

The client can already launch the browser once, copy the auto-generated token from the extension UI, and reuse it. That token is persisted across browser restarts, so subsequent runs don’t require any additional interaction. Functionally, this is equivalent.

The proposed alternatives also don’t remove user interaction:

- **Config file** — the extension can’t arbitrarily read files like `~/.playwright-mcp/...`; this would require explicit user action (e.g. a file picker or native messaging).
- **`chrome.storage.local`** — persistence across reinstalls doesn’t help with the initial bootstrap problem; you still need a browser session to set the token.
- **Environment variable** — extensions don’t have direct access to host environment variables without an external bridge, which again implies extra setup and user involvement.

So this still doesn’t solve the “zero browser interaction” automation case; it just moves the one-time interaction around.

yury-s:
Closing per above. Feel free to send another PR if you find a good solution.

PR #1328, opened 2026-01-23
Update README.md

Miridar:
Fixed extension README.md path after monorepo migration


pavelfeldman:
Ah, merged another one. Thanks though!

PR #1364, opened 2026-02-03
extension: support multiple simultaneous sessions

davidfant:
## What

Enable **multiple simultaneous MCP relay sessions** in the Chrome extension so different MCP clients can attach to different tabs at the same time.

## Why

The extension previously assumed a **single active connection**, which made parallel agent workflows impossible (or caused one session to kick the other). This change makes parallel sessions viable and improves UX by avoiding tab/window focus stealing.

## Tests (`packages/extension/tests/extension.spec.ts`)

* Launches a browser with the extension
* Opens two distinct pages/tabs
* Starts two MCP clients (`parallel-A`, `parallel-B`)
* Connects each client to a different tab via the extension connect UI
* Verifies isolation by checking that each client’s `browser_snapshot` returns the expected page content, including after running both snapshots in parallel

## Notes / Compatibility

* Backward compatible with existing UI consumers via `connectedTabId` (legacy field).
* Multi-session support is implemented primarily in the extension layer; the test is explicitly labeled a local spike to validate end-state behavior.

## How to test

1. Start two MCP clients using `--extension`
2. Connect each to a different tab from the connect UI
3. Confirm snapshots/actions remain isolated per client/tab

### DIFF
--- a/packages/extension/src/background.ts
+++ b/packages/extension/src/background.ts
@@ -30,12 +30,12 @@ type PageMessage = {
   type: 'getConnectionStatus';
 } | {
   type: 'disconnect';
+  tabId?: number;
 };
 
 class TabShareExtension {
-  private _activeConnection: RelayConnection | undefined;
-  private _connectedTabId: number | null = null;
   private _pendingTabSelection = new Map<number, { connection: RelayConnection, timerId?: number }>();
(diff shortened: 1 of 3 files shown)


davidfant:

pavelfeldman:
We are not ready to accept this contribution, but we'll think about it!

PR #1371, opened 2026-02-06
chore: mark v0.0.64

pavelfeldman:


packages/extension/tests/extension.spec.ts
```diff
@@ -400,7 +400,7 @@ test.describe('CLI with extension', () => {
     const confirmationPage = await confirmationPagePromise;
 
     // Click the Allow button
-    await confirmationPage.getByRole('button', { name: 'Allow' }).click();
+    await confirmationPage.getByRole('button', { name: 'Connect' }).first().click();
```
yury-s, line 403: let's fix the tool name in daemon

yury-s (approved):

packages/extension/tests/extension.spec.ts
pavelfeldman: It should probably go via the constructor not the tool name.

Issue #1100, opened 2025-10-01
browser_evaluate fails with "invalid JSON" error on large responses (potential Unicode serialization issue)


pinkwerks:
Description
browser_evaluate fails with "invalid JSON" when response contains Unicode surrogates

Summary
The browser_evaluate tool returns a 400 error when the evaluated JavaScript returns strings containing lone Unicode surrogates. The error manifests as:
API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"The request body is not valid JSON: no low surrogate in string: line 1 column 164755 (char 164754)"}}

Sketchy reproduction (couldn't get it to work all the time, but hits 100% of time in dev env)
This was the console errors returned from playwright

The server attempts to serialize invalid Unicode (lone surrogates) into JSON, which fails validation by the downstream API (Claude). I checked all of my source files for invalid low surrogates and came up empty.

Environment
OS: Windows 11
Playwright MCP version: [current version]
Browser: Chromium

<img width="1461" height="287" alt="Image" src="https://github.com/user-attachments/assets/c80ebaaf-6181-4c1a-b118-bde96ac18321" />

pinkwerks:
mcp__playwright__browser_wait_for

attached another log:

[console2.txt](https://github.com/user-attachments/files/22650697/console2.txt)

pinkwerks:
I was able to work around this by stripping all non-utf (emoji) from my console logs.

pavelfeldman:
Could you give me a couple of those so that I could put them into the console.log and add a test?

pinkwerks:
I'm in crunch mode at the moment, disabling my workaround to generate is going to be problematic for a bit. Also, I think the logs contain more info about my app that I would care to publish... to be absorbed as part of a test suite - so I hope you can understand that. Let me know when I can delete these logs, I'm not even sure if they trigger the bug for you?

pavelfeldman:
Go ahead and delete, we dan wait for the issue with the repro steps. I could not repro it with the emojis from your log anyways.

arthurflachs:
I am running into the same issue.

## How to reproduce

Page contains: `🦄`

Snapshot returns:
```yaml
- link "ðŸ¦„ ðŸ¦„" [ref=e3]:
  - generic [ref=e5]: ðŸ¦„
```

<details>
<summary>Test case that reproduces the issue</summary

```typescript
test('browser_snapshot handles emoji without breaking surrogate pairs', async ({ client, server }) => {
[...]

PR #1411, opened 2026-02-17
fix: sanitize lone Unicode surrogates in aria snapshots

### LINKED ISSUE #1410: browser_snapshot returns lone Unicode surrogates causing "no low surrogate in string" JSON errors
### Bug description

`browser_snapshot` (and potentially other tools returning page content) can return text containing lone Unicode surrogates from DOM `textContent`. These are valid in JavaScript strings but invalid in JSON (RFC 8259). When the MCP client serializes the tool result to JSON for the LLM API, it produces malformed JSON, causing a persistent 400 error:

```
API Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"The request body is not valid JSON: no low surrogate in string"}}
```

Once this content enters the conversation context, **every subsequent message fails** with the same error, making the session unrecoverable.

### Steps to reproduce

1. Navigate to any page containing emoji or special Unicode characters in its DOM (e.g., pages with emoji in text, or pages with content from user-generated sources)
2. Call `browser_snapshot`
3. The snapshot text may contain lone surrogates from the accessibility tree
4. Send the result to an LLM API — the JSON serialization fails

### Root cause

In the MCP response serialization path, the aria snapshot string is passed through without sanitizing Unicode surrogates. The snapshot comes from `page._snapshotForAI()` which reads DOM `textContent`. In JavaScript, strings can contain lone surrogates (U+D800-U+DFFF), but JSON requires surrogates to be properly paired (RFC 8259, Section 7). When `JSON.stringify()` encounters a lone surrogate, some implementations produce invalid JSON output.

### Suggested fix

Replace lone surrogates with U+FFFD (Unicode replacement character) before returning the snapshot text:

```js
// Strip unpaired surrogates (valid in JS strings, invalid in JSON)
function sanitizeSurrogates(str) {
  if (!str) return str;
  return str
    .replace(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/g, '\uFFFD')
    .replace(/(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g, '\uFFFD');
}
```

This preserves valid surrogate pairs (representing supplementary plane characters like emoji) while replacing only lone/unpaired surrogates.

### Related issues

- #1100 — same error, closed as "Not planned" due to lack of reproducible steps. This issue provides root cause analysis and a concrete fix.

### Environment

- Playwright MCP: 0.0.68
- Node.js: v22.x
- OS: Linux

alexmakeev:
## Summary

Lone Unicode surrogates (U+D800–U+DFFF) from DOM `textContent` are valid in JavaScript strings but invalid in JSON (RFC 8259 Section 7). When `browser_snapshot` returns text containing unpaired surrogates, JSON serialization fails with:

```
"no low surrogate in string"
```

This makes the MCP client session **unrecoverable** — every subsequent message fails with the same 400 error because the corrupted content sits in conversation context.

### Changes

- **Patch**: Replace lone surrogates with U+FFFD (replacement character) in aria snapshot text before returning from the MCP response serializer
- **Test**: Added test case that injects lone surrogates into DOM and verifies they are sanitized in the snapshot

### Root cause

`page._snapshotForAI()` reads DOM `textContent` which can contain lone surrogates (common with emoji, user-generated content, or encoding issues). The response serializer passes these through without sanitization, producing invalid JSON.

### Fix approach

Two-regex approach that preserves valid surrogate pairs while replacing only unpaired surrogates:
```js
str.replace(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/g, '\uFFFD')
   .replace(/(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g, '\uFFFD')
```

Fixes #1410
Related: #1100 (same error, closed as "Not planned" due to lack of reproducible steps)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,10 +7,14 @@
     "": {
       "name": "playwright-mcp-internal",
       "version": "0.0.68",
+      "hasInstallScript": true,
       "license": "Apache-2.0",
       "workspaces": [
         "packages/*"
       ],
+      "dependencies": {
+        "patch-package": "^8.0.0"
+      },
       "devDependencies": {
         "@modelcontextprotocol/sdk": "^1.25.2",
         "@playwright/test": "1.59.0-alpha-1771104257000",
@@ -1372,6 +1376,12 @@
         "vite": "^4.2.0 || ^5.0.0 || ^6.0.0 || ^7.0.0"
       }
     },
+    "node_modules/@yarnpkg/lockfile": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz",
+      "integrity": "sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==",
+      "license": "BSD-2-Clause"
+    },
     "node_modules/accepts": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-2.0.0.tgz",
@@ -1419,6 +1429,21 @@
         "ajv": {
           "optional": true
         }
+      }
+    },
+    "node_modules/ansi-styles": {
+      "version": "4.3.0",
+      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
+      "integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
+      "license": "MIT",
+      "dependencies": {
+        "color-convert": "^2.0.1"
+      },
+      "engines": {
+        "node": ">=8"
+      },
+      "funding": {
+        "url": "https://github.com/chalk/ansi-styles?sponsor=1"
       }
     },
     "node_modules/anymatch": {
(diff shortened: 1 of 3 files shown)


alexmakeev:

alexmakeev:
A note on reproducibility — I've verified this end-to-end with a red/green testing approach:

**RED phase (unpatched):** Running `browser_snapshot` on a page containing lone surrogates consistently produces:
```
API Error: 400 "no low surrogate in string: column NNNNN"
```
This error is fatal — it kills the entire MCP session, and every subsequent request also fails.

**GREEN phase (with this fix):** The same page snapshots successfully. Lone surrogates are replaced with U+FFFD (Unicode replacement character), preserving all other content intact.

**Reproduction materials:**
- A minimal self-contained HTML page is provided in [issue #1410](https://github.com/microsoft/playwright-mcp/issues/1410#issuecomment-3916939498) — it injects lone high surrogates (`\uD800`), lone low surrogates (`\uDC00`), and mixed strings into DOM `textContent` via JavaScript. This reliably triggers the error on any unpatched build.
- The PR includes a test that covers this exact scenario: inject surrogates → snapshot → verify sanitized output.

The bug surfaces on real-world pages (any site where JS assigns arbitrary text to `textContent`), so it's not a contrived edge case. I wanted to make sure the reproduction path is clear and the fix is verifiable in both directions.

pavelfeldman:
Closing as a low quality agentic patch, it patches node modules.

alexmakeev:
Following @pavelfeldman's suggestion, I've submitted a proper fix in the Playwright monorepo: https://github.com/microsoft/playwright/pull/39316

The fix is in the TypeScript source (`yaml.ts`), uses the same `toWellFormed()` pattern as `driver.ts`, and includes tests that pass in the monorepo test suite.

PR #1422, opened 2026-02-25
feat(extension): add orange tab group and isolate controlled tab in new window

mrw:
## Summary

I've been using Claude Cowork quite a bit, and when it takes over Chrome, it puts the tab it's controlling into a tab group with a color border. I've found that to be a really nice convenience feature because I can tell at quick glance which tabs are being automated. 

This PR adds that same feature to the Playwright MCP Bridge Chrome extension. It also opens the controlled tab in a new window.

- When the extension takes over a tab via the MCP bridge, it now moves the tab to a new browser window.
- The controlled tab is placed in an orange tab group labeled "Playwright MCP" in the browser tab strip.
- The tab group is automatically cleaned up when the MCP session disconnects or the tab is closed.

## Changes
| File | Change |
|---|---|
| `packages/extension/manifest.json` | Added `tabGroups` permission |
| `packages/extension/src/background.ts` | New window, orange tab group on connect, cleanup on disconnect |
| `packages/extension/tests/extension.spec.ts` | 3 tests: tab group creation, window isolation, cleanup on disconnect |

I ran the full test suite locally and they all passed.

### DIFF
--- a/packages/extension/tests/extension.spec.ts
+++ b/packages/extension/tests/extension.spec.ts
@@ -187,6 +187,13 @@ async function startWithExtensionFlag(browserWithExtension: BrowserWithExtension
   return client;
 }
 
+async function getServiceWorker(context: BrowserContext) {
+  let [sw] = context.serviceWorkers();
+  if (!sw)
+    sw = await context.waitForEvent('serviceworker');
+  return sw;
+}
+
 const testWithOldExtensionVersion = test.extend({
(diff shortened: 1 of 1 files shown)


mrw:
@microsoft-github-policy-service agree company="Happy Cog"

yury-s:
Implemented in https://github.com/microsoft/playwright-mcp/pull/1514

PR #1448, opened 2026-03-06
fix: sanitize malformed Unicode in MCP responses

### LINKED ISSUE #1447: Fix malformed Unicode in outgoing MCP responses
A browser action can succeed, but the follow-up MCP response fails to serialize with:

	the request body is not valid JSON: invalid high surrogate in string

This appears to happen when tool output includes malformed Unicode from page-derived text.

## Reproduction

- Start `@playwright/mcp` with `--extension`
- Navigate to a page containing problematic text content
- Perform a browser action such as `browser_click`
- Observe that the action succeeds, but the returned MCP response fails

## Observed behavior

The browser action completes, but the client receives an error during response handling:

`invalid high surrogate in string`

## Expected behavior

If page-derived text contains malformed Unicode, MCP should still return a valid JSON-RPC response instead of failing serialization.

## Likely root cause

The outgoing MCP message contains a JavaScript string with a lone surrogate or other non-well-formed Unicode sequence. This is most likely coming from page-derived strings such as accessibility snapshot text, console text, page title, URL, or other extracted content.

## Evidence

A local wrapper that sanitizes outgoing JSON-RPC string values before transport serialization fixes the issue without changing browser behavior. That strongly suggests the bug is in outgoing message serialization, not in the browser action itself.

## Suggested fix

Sanitize outgoing MCP message strings before transport serialization:

- use `String.prototype.toWellFormed()` when available
- otherwise replace lone surrogates with `\uFFFD`

Apply this at the transport boundary so it covers stdio, SSE, and streamable HTTP responses.

frankhommers:
## Summary
- sanitize malformed Unicode just before MCP transport sends responses over stdio, SSE, and streamable HTTP
- preserve non-plain values that already serialize correctly while repairing lone surrogate text content
- add regression coverage for malformed Unicode transport output and non-plain value preservation

## Testing
- `npx playwright test tests/unicode-serialization.spec.ts`
- `node packages/playwright-mcp/cli.js --help`

Closes #1447

### DIFF
--- a/packages/playwright-mcp/cli.js
+++ b/packages/playwright-mcp/cli.js
@@ -16,9 +16,81 @@
  */
 
 const { program } = require('playwright-core/lib/utilsBundle');
+const mcpBundle = require('playwright-core/lib/mcpBundle');
 const { decorateMCPCommand } = require('playwright/lib/mcp/program');
 
 const packageJSON = require('./package.json');
-const p = program.version('Version ' + packageJSON.version).name('Playwright MCP');
-decorateMCPCommand(p, packageJSON.version)
-void program.parseAsync(process.argv);
+
+function sanitizeString(value) {
+  if (typeof value !== 'string')
+    return value;
+  if (typeof value.toWellFormed === 'function')
+    return value.toWellFormed();
+
+  let result = '';
+  for (let index = 0; index < value.length; index++) {
+    const code = value.charCodeAt(index);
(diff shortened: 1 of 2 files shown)


frankhommers:

pavelfeldman:
That's a very random place for a fix.

frankhommers:
@pavelfeldman Fair enough! Could you point me in the right direction for where this fix should live instead? Happy to reopen in the correct place once I know where that would be.

PR #1451, opened 2026-03-10
docs: add Chinese translation to README

hiSandog:
Add Chinese localization to help Chinese-speaking users better understand and use Playwright MCP.


yury-s:
I don't think we want this, modern tools can translate it very well in place.

PR #1468, opened 2026-03-16
Update environment variable for extension connection

jeremytenjo:
image


yury-s:
`PLAYWRIGHT_MCP_EXTENSION_TOKEN` is actually a different env variable that doesn't have a counterpart cli option. So the change is incorrect.

PR #1471, opened 2026-03-18
feat(extension): add focusTab option to prevent focus stealing on connect

himself65:
## Summary
- Adds a `focusTab` URL parameter to the extension connect page that controls whether connecting to a tab activates it and focuses its window.
- When `focusTab=false` is passed, the extension skips calling `chrome.tabs.update(tabId, { active: true })` and `chrome.windows.update(windowId, { focused: true })`, preventing focus from being stolen from the terminal.
- Defaults to `true` for backward compatibility.

### DIFF
--- a/packages/extension/src/background.ts
+++ b/packages/extension/src/background.ts
@@ -26,6 +26,7 @@ type PageMessage = {
   tabId?: number;
   windowId?: number;
   mcpRelayUrl: string;
+  focusTab?: boolean;
 } | {
   type: 'getConnectionStatus';
 } | {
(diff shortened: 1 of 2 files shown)


himself65:

yury-s:
Can you describe a scenario where this is needed?

himself65:
> Can you describe a scenario where this is needed?

Yes, im running agent on background and I dont want to focus the window

yury-s:
> > Can you describe a scenario where this is needed?
> 
> Yes, im running agent on background and I dont want to focus the window

In that case you could just pass a token and skip the tab selection dialog, for that scenario we could avoid activating the tab. If the agent connects and the extension shows tab selection dialog, and the user clicks `connect`, why would you want to suppress switching to that tab?

Can you describe full scenario as the patch also doesn't have any call sites that would pass `focusTab=false`?

PR #1478, opened 2026-03-20
fix(extension): support multiple tabs opened by Playwright

### LINKED ISSUE #1317: Allow the MCP server (with --extension) to manage multiple tabs
It seems not possible to manage multiple tabs using the MCP server and the bridge extension.

snomiao:
Fixes #1317

## Problem

When using \`--extension\` mode, Playwright can only control the single tab that was manually selected in the extension popup. Any tab Playwright opens via \`tab-new\` (i.e. \`Target.createTarget\`) is created but immediately uncontrollable.

**Root cause:** \`_onDebuggerEvent\` in \`RelayConnection\` only forwards CDP events where \`source.tabId === this._debuggee.tabId\`. New tabs never have a debugger attached and their events are never forwarded — so Playwright sees them in \`tab-list\` but cannot interact with them.

### \`relayConnection.ts\`
- Add \`createTab\` command: creates a Chrome tab via \`chrome.tabs.create()\`, attaches the debugger, and returns \`{ tabId, targetInfo }\` to the relay
- Track Playwright-opened tabs in \`_playwrightTabIds: Set<number>\`
- Forward CDP events from all Playwright-opened tabs (tagged with \`tabId\` for relay routing)
- Route \`forwardCDPCommand\` to the correct debuggee by \`tabId\`
- \`attachToTab\` now returns \`tabId\` so the relay can build its session→tab mapping
- Add \`onPlaywrightTabCreated\` / \`onPlaywrightTabRemoved\` callbacks for the background script
- Detach debugger from all Playwright tabs on close

### \`background.ts\`
- Replace single \`_activeConnection\` with \`_connections: Map<mcpRelayUrl, ConnectionState>\` — each relay URL (UUID-based) gets fully isolated state
- Multiple simultaneous MCP instances no longer conflict in the extension
- Track Playwright tab IDs per connection; clear badges and the set when the connection closes
- Show a blue \`✓\` badge on Playwright-managed tabs (distinct from the green \`✓\` on the initial connected tab)
- Expose \`playwrightTabIds\` via \`getConnectionStatus\`
- Clean up Playwright tab IDs in \`_onTabRemoved\`

### \`status.tsx\`
- Show all active connections, each with their own Playwright-managed tab list
- Add a "Playwright managed tabs" section per connection, clickable to focus

## Multi-client architecture

With \`--port\` (HTTP mode), a single playwright-mcp process handles multiple clients via HTTP Streamable (\`POST /mcp\`) or SSE (\`GET /sse\`). Each MCP session calls \`createExtensionBrowser()\`, which creates a new \`CDPRelayServer\` with a unique UUID relay URL.

Before this PR, the extension only tracked one relay connection — the second client would be rejected. The multi-instance fix in \`background.ts\` makes the extension handle each relay URL independently:

\`\`\`
playwright-mcp --extension --port 4321   (one shared process)
  ├── Claude session A  →  CDPRelayServer (uuid-aaa)  →  extension ConnectionState A  →  Tab 1
  └── Claude session B  →  CDPRelayServer (uuid-bbb)  →  extension ConnectionState B  →  Tab 2
\`\`\`

Each session needs its own browser tab (Chrome only allows one debugger per tab). As long as each client connects to a different tab, they are fully isolated with no conflicts.

The companion relay-server changes (protocol + \`cdpRelay.ts\`) are in microsoft/playwright#39805.

### DIFF
--- a/packages/extension/src/background.ts
+++ b/packages/extension/src/background.ts
@@ -35,6 +35,7 @@ type PageMessage = {
 class TabShareExtension {
   private _activeConnection: RelayConnection | undefined;
   private _connectedTabId: number | null = null;
+  private _playwrightTabIds: Set<number> = new Set();
   private _pendingTabSelection = new Map<number, { connection: RelayConnection, timerId?: number }>();
 
   constructor() {
@@ -67,7 +68,8 @@ class TabShareExtension {
         return true; // Return true to indicate that the response will be sent asynchronously
       case 'getConnectionStatus':
         sendResponse({
-          connectedTabId: this._connectedTabId
+          connectedTabId: this._connectedTabId,
+          playwrightTabIds: [...this._playwrightTabIds],
         });
         return false;
       case 'disconnect':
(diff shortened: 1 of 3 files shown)


snomiao:

pavelfeldman:
We are not ready to accept it sorry.

PR #1532, opened 2026-04-08
docs: replace per-host install blocks with agent-add one-command install

pea3nut:
## Summary

Replaces the 8 per-host `<details>` install blocks (VS Code / Cursor / Windsurf / Claude Code / Claude Desktop / Goose / Qodo Gen / LM Studio) with a single [agent-add](https://github.com/pea3nut/agent-get) one-command install, plus a minimal manual-JSON fallback.

```bash
npx -y agent-add --mcp '{"playwright":{"command":"npx","args":["@playwright/mcp@latest"]}}'
```

`agent-add` auto-detects the user's AI host (18 supported) and writes the MCP config to the right location, so the per-host blocks become redundant.

Docs-only — no code changes. The manual-JSON fallback is preserved for users who prefer to edit their config by hand.


pea3nut:

PR #1536, opened 2026-04-09
Honor context options for isolated CDP sessions

tonyespinoza1:
## Summary
- pass `browserConfig.contextOptions` when creating isolated CDP contexts
- add a regression test covering `--cdp-endpoint + --isolated + --storage-state`

## Why
`--storage-state` and other context options are already respected for launched isolated contexts, but CDP-attached isolated contexts were calling `browser.newContext()` without the configured options. That makes `--isolated` behave inconsistently between launch and CDP modes.

## Proof
- added `cdp server honors storage state for isolated contexts` in `tests/cdp.spec.ts`
- ran:
  - `npx playwright test tests/cdp.spec.ts --project=chrome -g 'cdp server honors storage state for isolated contexts'`

### DIFF
--- a/src/browserContextFactory.ts
+++ b/src/browserContextFactory.ts
@@ -125,7 +125,7 @@ class CdpContextFactory extends BaseContextFactory {
   }
 
   protected override async _doCreateContext(browser: playwright.Browser): Promise<playwright.BrowserContext> {
-    return this.browserConfig.isolated ? await browser.newContext() : browser.contexts()[0];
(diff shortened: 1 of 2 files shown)


tonyespinoza1:
@microsoft-github-policy-service agree company="CommonToolsInc"

tonyespinoza1:
Current upstream main already passes the CDP isolated storage-state regression this PR was meant to fix. The repo has also been refactored since this was opened, so the original patch no longer applies cleanly. I verified the behavior by running an equivalent focused regression on current main, and it passes there, so I’m closing this as obsolete rather than forcing a stale patch through.

PR #1561, opened 2026-04-19
docs: note chromium setup for linux arm64

### LINKED ISSUE #1560: Linux Arm64 - Chrome Not available. Use chromium
I am unsure how to go about this but as Chrome is not available on Linux ARM64 I believe Chromium should be used instead. It seems as if the MCP will not work though. Is there a way to have it work? Everytime I try to use it, it shows Chrome not installed. I try instructing the agent to use Chromium instead however at that point, it bypasses the MCP entirely.

MukundaKatta:
## Summary
- document that Linux arm64 users may need to select chromium explicitly
- add a ready-to-copy MCP config example using --browser chromium

## Testing
- not run (README-only change)

Closes #1560

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,24 @@ First, install the Playwright MCP server with your client.
 }
 ```
 
+If you're on Linux arm64, the Chrome channel may be unavailable. In that case,
+configure Playwright MCP to use `chromium` explicitly instead of `chrome`:
+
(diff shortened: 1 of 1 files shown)


pavelfeldman:
Linux arm users are very savvy, they know what they are doing!

PR #1578, opened 2026-04-28
Clean up cliDaemon processes on MCP server exit

akimovmike:
## Problem

When the MCP wrapper (`@playwright/mcp`) is killed by an MCP client — including via SIGKILL or simple stdio close — the `cliDaemon` it spawned stays alive forever, along with its headless browser. The daemon is reparented to init (PPID=1) and continues consuming RAM.

**Repro:**

1. Start any MCP client that talks to `npx @playwright/mcp@latest` over stdio.
2. Issue a tool call that opens a browser (e.g. `browser_navigate`).
3. Kill the wrapper: `pkill -9 -f '@playwright/mcp'`.
4. Observe: `pgrep -fa cliDaemon.js` still alive with `PPID=1`, plus headless Chrome.

I've personally hit cliDaemons surviving 2+ days after their MCP parent died, accumulating ~75 MB RAM each across long-lived shell sessions running multiple MCP clients.

## Root cause

`playwright-core/lib/tools/cli-client/session.js`'s `Session.startDaemon()` spawns the daemon with `detached: true` + `child.unref()`. SIGINT/SIGTERM forwarding handlers are removed once the daemon is up. From the wrapper's perspective, the daemon is intentionally orphaned for CLI persistence — but in MCP mode that lifecycle is wrong: when the MCP server exits, the daemon should go too.

## Fix

`cli.js` snapshots existing cliDaemon PIDs at startup, then registers signal/exit/stdin-EOF handlers that kill any *new* cliDaemon PIDs that appeared during this server's lifetime.

- **Diff-based snapshot** — never kills daemons started by other concurrent MCP servers, only ours.
- **Signal handlers** — SIGINT, SIGTERM, SIGHUP run blocking cleanup (SIGTERM → 500ms wait → SIGKILL).
- **Stdin EOF** — primary disconnect signal for MCP stdio transport. Skipped in HTTP mode (`--port`/`--host`).
- **Sync `exit` handler** — last-ditch fire-and-forget SIGTERM for paths where signals don't run.
- **Windows** — no-op (no PPID/pgid model in the same form).

## Why not fix in `playwright-core`

The orphan-on-MCP-exit behavior is specific to the MCP wrapper's lifecycle expectations. CLI users explicitly rely on the daemon outliving short-lived `pw cli ...` invocations. Adding cleanup logic here keeps the change scoped to the MCP server entry point and doesn't risk regressing CLI flows.

## Verification

```bash
# Build
node cli.js --version  # → "Version 0.0.71"

# Smoke test: stdin EOF triggers shutdown with exit 143 (SIGTERM)
node -e "const cp=require('child_process'); const p=cp.spawn('node',['./cli.js']); setTimeout(()=>p.stdin.end(),500); p.on('exit',(c,s)=>console.log(c,s));"
# → 143 null
```

I'd happily add a Playwright-test integration spec under `tests/` that opens a daemon via an MCP tool call, closes stdin, then asserts `pgrep` no longer finds the daemon — let me know if that's wanted as part of this PR or a follow-up.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+const { spawnSync } = require('child_process');
 const { program } = require('playwright-core/lib/utilsBundle');
 const { tools, libCli } = require('playwright-core/lib/coreBundle');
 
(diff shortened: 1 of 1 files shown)


pavelfeldman:
cliDaemon is not used by mcp

PR #1602, opened 2026-05-05
docs: Fixed secrets env variable name

grulex:
Incorrect name PLAYWRIGHT_MCP_SECRETS replaced to correct PLAYWRIGHT_MCP_SECRETS_FILE


grulex:

grulex:
reference https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/tools/mcp/config.ts#L405

pavelfeldman (approved):

PR #1622, opened 2026-05-12
test(capabilities): verify opt-in tools absent from default snapshot list

shineemad:
- Add negative test to verify that opt-in tools (`browser_pdf_save`, `browser_mouse_*`) are not present in the default snapshot tool list

### DIFF
--- a/tests/capabilities.spec.ts
+++ b/tests/capabilities.spec.ts
@@ -14,64 +14,77 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


yury-s:
Is there an issue it is fixing? If so, please start with filing one before sending a pr. Also bulk of mcp tests live in https://github.com/microsoft/playwright, only small subset is here to validate rolls.

PR #1623, opened 2026-05-15
docs: document Windows Claude Code workaround

MukundaKatta:
## Summary
- Document Windows Claude Code workaround

## Changed files
- `README.md`

## Tests
- Not run in this publishing pass.


yury-s:
We should fix the underlying bug if there is one, rather than suggesting this. Please file an issue with repro steps if you face a problem with launching playwright mcp.

PR #1699, opened 2026-07-29
fix(security): require a bearer token instead of shipping an open server

Ho1yShif:
## Summary
- Adds `render-auth-proxy.mjs` (Node stdlib only): binds `$PORT`, answers `401` to anything without `Authorization: Bearer $MCP_TOKEN`, forwards the rest to the MCP server — which now binds to `127.0.0.1:8931` and is never published.
- Rate-limits token guessing: after 10 failed attempts in a minute a client gets `429` with `Retry-After`, and the log gets one line at the threshold rather than one per attempt. The limit is scoped to requests that were already going to be refused, so a valid token is never locked out and nobody can aim a lockout at the operator.
- `render.yaml` generates `MCP_TOKEN` per deploy (`generateValue: true`); the container refuses to start if it's unset, so there is no unauthenticated configuration and no shared secret in the repo.
- Adds `render-smoke-test.sh` and points CI at it: builds the image and drives the real container end to end. This template wraps a prebuilt image rather than building the server, so the wrapper is the only thing it owns and the only thing worth gating.
- README: deploy step for copying the token, client config with the header, rewritten Security section (what the gate does, what it doesn't), plus `.env.example` and roll-checklist updates.

## Why
Upstream ships no auth in HTTP mode and exposes `browser_run_code_unsafe`, which `--caps` cannot remove. That's defensible for `npx` on loopback; as a one-click Render template it meant a public URL was the only thing between the internet and code execution in the forker's container — and a pivot into their private network. A README warning isn't a control.

## Notes
- Constant-time comparison over SHA-256 digests; the presented token is never logged.
- Client identity for the rate limiter is the address Render's edge observed (rightmost `X-Forwarded-For`). Off Render nothing trusted sets that header, so it is ignored entirely and the socket address is used. The failure map is swept per window and capped, so a source-rotating attacker can't grow it without bound.
- `Host` is forwarded unchanged, so upstream's `--allowed-hosts` DNS-rebinding check still behaves as documented.
- The proxy is PID 1: it supervises the server, exits with its code, and forwards `SIGTERM` so Render shutdowns stay clean. It also holds `$PORT` closed until upstream accepts, so Render's open-port liveness signal means the whole stack is up rather than just the gate.

## Verification
Blueprint: `render blueprints validate render.yaml` → `valid: true`.

`./render-smoke-test.sh` passes against the built container — fails closed with no `MCP_TOKEN`, `401` without the header, exactly 10 `401`s then `429` for repeated bad tokens with the valid token still returning `200`, an MCP handshake, a real `browser_navigate` round-tripping through the SSE path, and `SIGTERM` stopping it in 0s with exit 0. CI runs the same script on every PR.

Proxy also verified directly against a stub upstream (valid-token prefix and `Basic` instead of `Bearer` → `401` with `WWW-Authenticate: Bearer`; body and `Host` forwarded intact; SSE streamed incrementally; child exit code propagated; no orphan after `SIGTERM`; upstream down → `502`, no hang; no token value anywhere in the logs).

**Not yet done:** a real fresh-fork deploy. Needs that green before merge.


Ho1yShif:
Opened against the wrong repository by mistake — this change belongs to the render-examples template fork, not upstream. Closing; apologies for the noise.

Issue #1736, opened 2026-09-05
Proposal: add an Agent Plugins 1.0 package


777genius:
## Problem

Playwright MCP currently documents separate setup paths for each client. [Agent Plugins 1.0](https://agent-plugins.org/specification) defines a portable package boundary that can carry the same MCP configuration across compatible clients.

## Proposed scope

I would like to add a small official package with:

- root `plugin.json` and `mcp.json` files
- an exact `@playwright/mcp` version in `mcp.json`, updated by the existing release flow
- no runtime code, dependency, command behavior, or default behavior changes
- a short optional README example: `npx universal-agent-plugins add playwright`

The package would remain useful without that installer because compatible clients can read the standard files directly.

## Preflight evidence

I prepared the package locally before opening this issue:

- Agent Plugins 1.0 schema validation passes
- add, info, update, and remove pass in isolated synthetic profiles for Codex, Cursor, Kiro, Gemini CLI, OpenCode, Cline, Windsurf, and VS Code
- the exact `@playwright/mcp@0.0.80` command initializes through MCP Inspector and returns 24 tools
- no browser action, credential, or real user profile was used

Per the contribution policy, I am asking before submitting the change. If this scope fits the project, may I prepare the PR and be assigned to this issue?

777genius:
Update: the reviewed short alias is now live and reproducible:

```bash
npx universal-agent-plugins add playwright
```

The exact published package passed `add`, `info`, no-change `update`, and purge `remove` in fresh credential-free roots for all ten supported client adapters: Codex, Cursor, Copilot CLI, VS Code, Kiro, Claude Code, Gemini CLI, OpenCode, Cline, and Windsurf. The run left zero installations, open operations, or owned artifacts: https://github.com/777genius/universal-agent-plugins-registry/actions/runs/33939289307

Package page: https://777genius.github.io/universal-agent-plugins-registry/plugins/playwright

No browser action, account credential, or real user profile was used. I am still waiting for assignment or approval before opening the PR, per the contribution policy.

PR #1746, opened 2026-09-11
feat: add Agent Plugins 1.0 package

pavelfeldman:
## Summary
- Add `plugin.json` and `mcp.json` at the repository root so clients that support [Agent Plugins 1.0](https://agent-plugins.org) (VS Code, Copilot CLI, Cursor, Codex, Kiro) can install the server from the repository URL
- `mcp.json` pins the exact `@playwright/mcp` version; the publish workflow validates it and `plugin.json` against `package.json` alongside the existing `server.json` check

Fixes https://github.com/microsoft/playwright-mcp/issues/1736

### DIFF
--- a/mcp.json
+++ b/mcp.json
@@ -1,0 +1,12 @@
+{
+  "$schema": "https://agent-plugins.org/schemas/1.0.0/mcp.schema.json",
+  "mcpServers": {
+    "playwright": {
+      "type": "stdio",
+      "command": "npx",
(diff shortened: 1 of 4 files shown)


yury-s (approved):

mcp.json
yury-s, line 1: I think these files should probably go under `plugin/` or something which eventually could also get skills.

